PR #1, opened 2022-08-31
Ingress interface standard

jdkandersson:
- Minor styling improvements
- Fix a broken unit test
- Add a unit test for the new interface standard and fix a bug it uncovered

### DIFF
--- a/lib/charms/nginx_ingress_integrator/v0/ingress.py
+++ b/lib/charms/nginx_ingress_integrator/v0/ingress.py
@@ -125,9 +125,7 @@ class IngressRequires(Object):
     def __init__(self, charm, config_dict):
         super().__init__(charm, "ingress")
 
-        self.framework.observe(
-            charm.on["ingress"].relation_changed, self._on_relation_changed
-        )
+        self.framework.observe(charm.on["ingress"].relation_changed, self._on_relation_changed)
 
         # Set default values.
         DEFAULT_RELATION_FIELDS = {
@@ -163,9 +161,7 @@ class IngressRequires(Object):
             self.model.unit.status = BlockedStatus(blocked_message)
             return True
         if not update_only:
-            missing = [
-                x for x in REQUIRED_INGRESS_RELATION_FIELDS if x not in self.config_dict
-            ]
+            missing = [x for x in REQUIRED_INGRESS_RELATION_FIELDS if x not in self.config_dict]
             if missing:
                 LOGGER.error(
                     "Ingress relation error, missing required key(s) in config dictionary: %s",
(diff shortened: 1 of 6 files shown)


lib/charms/nginx_ingress_integrator/v0/ingress.py
```diff
@@ -117,7 +125,22 @@ class IngressRequires(Object):
...
-        self.framework.observe(charm.on["ingress"].relation_changed, self._on_relation_changed)
+        self.framework.observe(
+            charm.on["ingress"].relation_changed, self._on_relation_changed
+        )
+
+        # Set default values.
+        DEFAULT_RELATION_FIELDS = {
+            "service-namespace": self.model.name,
+        }
+        for k, v in DEFAULT_RELATION_FIELDS.items():
```
arturo-seijas, line 136: Aren't the two statements in this if clause the same?

lib/charms/nginx_ingress_integrator/v0/ingress.py
```diff
@@ -117,7 +125,22 @@ class IngressRequires(Object):
...
+            if k not in config_dict or not config_dict[k]:
+                config_dict[k] = v
+
+        # And now populate data for conformity with charm-relation-interfaces.
+        for k, v in RELATION_INTERFACES_MAPPINGS.items():
+            if k in config_dict and config_dict[k]:
```
arturo-seijas, line 142: Do we need to check the existence of the key given this is validated afterwards in the `_config_dict_errors` method?

arturo-seijas (changes requested):

mthaddon (commented):
There seem to be quite a few formatting changes here. Can we remove those to make the diff easier to read, just including things that have changed (running `make blacken` doesn't suggest any formatting changes are needed).

As an example you're splitting this onto three lines even though it's less than the maximum line length:
`if field in BOOLEAN_CONFIG_FIELDS and (config_data is not None and config_data != ""):`

jdkandersson:
> 



> There seem to be quite a few formatting changes here. Can we remove those to make the diff easier to read, just including things that have changed (running `make blacken` doesn't suggest any formatting changes are needed)?
> 
> As an example you're splitting this onto three lines even though it's less than the maximum line length: `if field in BOOLEAN_CONFIG_FIELDS and (config_data is not None and config_data != ""):`

I'll check, those were all done automatically by the code editor by running black

jdkandersson:
> 



> > There seem to be quite a few formatting changes here. Can we remove those to make the diff easier to read, just including things that have changed (running `make blacken` doesn't suggest any formatting changes are needed)?
> > As an example you're splitting this onto three lines even though it's less than the maximum line length: `if field in BOOLEAN_CONFIG_FIELDS and (config_data is not None and config_data != ""):`
> 
> I'll check, those were all done automatically by the code editor by running black

So what happened is that the `make blacken` command used a line length of 99 as a parameter. The code editor used the default 80 and reformatted a bunch of code. I have now added the line length to `pyproject.toml` which means this shouldn't happen again.

lib/charms/nginx_ingress_integrator/v0/ingress.py
jdkandersson: I don't quite follow, this step is performing a mapping of the old keys to the new keys defined in the ingress standard to ensure compatibility with both. The code isn't very clear, I'll try and update it to make it clearer.

lib/charms/nginx_ingress_integrator/v0/ingress.py
arturo-seijas: Oh, yes. Makes sense. Nvm

arturo-seijas (approved):
LGTM

PR #7, opened 2022-09-02
Ingress requires unit tests

jdkandersson:
Add unit tests for `IngressRequires`

### DIFF
--- a/lib/charms/nginx_ingress_integrator/v0/ingress.py
+++ b/lib/charms/nginx_ingress_integrator/v0/ingress.py
@@ -142,15 +142,9 @@ class IngressRequires(Object):
         """Check our config dict for errors."""
         blocked_message = "Error in ingress relation, check `juju debug-log`"
         unknown = [
-<<<<<<< HEAD
-            x
-            for x in self.config_dict
-            if x
-=======
             config_key
             for config_key in self.config_dict
             if config_key
->>>>>>> ingress-interface-standard-ingress-requires-unit-tests
             not in REQUIRED_INGRESS_RELATION_FIELDS
             | OPTIONAL_INGRESS_RELATION_FIELDS
             | RELATION_INTERFACES_MAPPINGS_VALUES
--- a/lib/charms/nginx_ingress_integrator/v0/ingress.py
+++ b/lib/charms/nginx_ingress_integrator/v0/ingress.py
@@ -162,15 +162,6 @@ class IngressRequires(Object):
                 for config_key in REQUIRED_INGRESS_RELATION_FIELDS
                 if config_key not in self.config_dict
             ]
-            # Mapp configuration keys to new keys
-            missing = list(
-                map(
-                    lambda config_key: RELATION_INTERFACES_MAPPINGS[config_key]
-                    if config_key in RELATION_INTERFACES_MAPPINGS
-                    else config_key,
-                    missing,
-                )
-            )
             if missing:
                 LOGGER.error(
                     "Ingress relation error, missing required key(s) in config dictionary: %s",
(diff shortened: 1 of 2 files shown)


weiiwang01 (dismissed):
LGTM, thanks!

merkata (dismissed):
Looks really good! I see we still don't have CI/CD for this charm in place and am wondering whether we shouldn't add it?

lib/charms/nginx_ingress_integrator/v0/ingress.py
```diff
@@ -161,7 +157,20 @@ def _config_dict_errors(self, update_only=False):
             self.model.unit.status = BlockedStatus(blocked_message)
             return True
         if not update_only:
-            missing = [x for x in REQUIRED_INGRESS_RELATION_FIELDS if x not in self.config_dict]
+            missing = [
+                config_key
+                for config_key in REQUIRED_INGRESS_RELATION_FIELDS
+                if config_key not in self.config_dict
+            ]
+            # Mapp configuration keys to new keys
```
mthaddon, line 165: s/Mapp/Map/ ?

tests/unit/test_library_ingress.py
```diff
@@ -1,50 +1,216 @@
...
+
+        self.assertEqual(self.harness.charm.ingress.config_dict["host"], service_hostname)
+        self.assertEqual(self.harness.charm.ingress.config_dict["name"], service_name)
+        self.assertEqual(self.harness.charm.ingress.config_dict["port"], service_port)
+        self.assertEqual(self.harness.charm.ingress.config_dict["model"], service_namespace)
+
+    def test_with_new_config(self):
+        """
+        arrange: given a configuration with all the items pre-standard
+        act: when the charm is constructed with the configuration
```
mthaddon, line 116: Should these two lines be different from the ones in `test_with_previous_config`?

lib/charms/nginx_ingress_integrator/v0/ingress.py
```diff
@@ -161,7 +157,20 @@ def _config_dict_errors(self, update_only=False):
...
+            missing = list(
+                map(
+                    lambda config_key: RELATION_INTERFACES_MAPPINGS[config_key]
+                    if config_key in RELATION_INTERFACES_MAPPINGS
+                    else config_key,
+                    missing,
+                )
+            )
```
mthaddon, line 173: I think this is going to mean that we're returning the new key names as what's missing, but the charm library as currently written still requires the old values, so that's going to be a bit confusing.

What's changing is what's being passed over the relation, not what users are asked to use when working with this charm library.

mthaddon (changes requested):
I'm not sure we should be returning the new values as part of the log message, since the charm author is still defining the config_dict using service-hostname, service-name and service-port in the charm library.

jdkandersson:
> Looks really good! I see we still don't have CI/CD for this charm in place and am wondering whether we shouldn't add it?

Yes we should, this charm has previously been cited as lower priority, I'll add a Jira story for the CI/CD for it

lib/charms/nginx_ingress_integrator/v0/ingress.py
jdkandersson: Thank you, fixed in next commit

tests/unit/test_library_ingress.py
jdkandersson: Thank you, fixed in next commit

lib/charms/nginx_ingress_integrator/v0/ingress.py
jdkandersson: Thanks, fixed

weiiwang01 (approved):
Everything seems to be good, but can we summarize the git commit message subject down to less than 50 characters more often so Github wouldn't clip the message.

mthaddon (approved):
Very nice, thanks for these changes!

PR #9, opened 2022-09-06
Integration test

jdkandersson:
Adds 2 integration tests:

- That the workload becomes active
- That it is possible to communicate with another charm through the ingress relation

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -1,0 +1,52 @@
+# Copyright 2022 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+from pathlib import Path
+
+import pytest
+import pytest_asyncio
+import yaml
+from ops.model import ActiveStatus
+from pytest import fixture
+from pytest_operator.plugin import OpsTest
+
+
+@fixture(scope="module")
+def metadata():
+    """Provides charm metadata."""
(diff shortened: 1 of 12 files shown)


tests/integration/conftest.py
```diff
@@ -0,0 +1,53 @@
...
+    # Check that both ingress and hello-kubecon are active
+    assert ops_test.model.applications[app_name].units[0].workload_status == ActiveStatus.name
+    assert (
+        ops_test.model.applications[hello_kubecon_app_name].units[0].workload_status
+        == ActiveStatus.name
+    )
+
+    # Add required relations
+    await ops_test.model.add_relation(hello_kubecon_app_name, app_name)
+    await ops_test.model.wait_for_idle(status="active")
```
arturo-seijas, line 51: Maybe it'd be better to use ActiveStatus.name instead of "active"

arturo-seijas (changes requested):

weiiwang01 (approved):

tests/integration/conftest.py
jdkandersson: Thanks for pointing this out, fixed

weiiwang01 (approved):
LGTM, thanks!

arturo-seijas (approved):
LGTM

PR #12, opened 2022-09-14
Increment patch level so a new library version can be published

mthaddon:
In the previous change we updated the docstring for the charm library to fix formatting of lists. I had mistakenly thought that when you published a new version of the charm that library would be updated as well. It turns out you need to run `charmcraft publish-lib` specifically, so I'm incrementing the patch number here to allow me to do that.


arturo-seijas (approved):
LGTM

jdkandersson (approved):

merkata (approved):

Issue #13, opened 2022-09-29
Add Whitelist Source Range Config Option


kian99:
Hi,

A useful config addition would be the ability to add a range of whitelisted addresses as described here - https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#whitelist-source-range

Thanks,
Kian

mthaddon:
This has been added, and tested as follows https://paste.ubuntu.com/p/vrGtWtwj74/.

Released as revision 29 in the stable channel.

PR #14, opened 2022-09-29
Add whitelist-source-range annotation based on config entry

mthaddon:
Addresses https://github.com/canonical/nginx-ingress-integrator-operator/issues/13.

I believe this would only ever be a deploy-time option, so it's not been exposed as an option in the relation. That could obviously be changed in the future though.

### DIFF
--- a/tests/unit/test_charm.py
+++ b/tests/unit/test_charm.py
@@ -446,14 +446,12 @@ class TestCharm(unittest.TestCase):
         conf_or_rel = self.harness.charm._all_config_or_relations[0]
         self.assertEqual(conf_or_rel._whitelist_source_range, "10.0.0.0/24,172.10.0.1")
         result_dict = conf_or_rel._get_k8s_ingress().to_dict()
-        # Expect default values plus the one we've added.
-        expected = {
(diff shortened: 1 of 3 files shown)


tests/unit/test_charm.py
```diff
@@ -441,6 +441,20 @@ def test_rewrite_enabled_property(self):
         conf_or_rel = self.harness.charm._all_config_or_relations[0]
         self.assertEqual(conf_or_rel._rewrite_enabled, False)
 
+    def test_whitelist_source_range(self):
+        self.harness.update_config({"whitelist-source-range": "10.0.0.0/24,172.10.0.1"})
+        conf_or_rel = self.harness.charm._all_config_or_relations[0]
+        self.assertEqual(conf_or_rel._whitelist_source_range, "10.0.0.0/24,172.10.0.1")
+        result_dict = conf_or_rel._get_k8s_ingress().to_dict()
+        # Expect default values plus the one we've added.
+        expected = {
```
arturo-seijas, line 450: I wouldn't compare the rest of the properties values in case they are changes. It'll make the tests hard to maintain

arturo-seijas (approved):
LGTM

weiiwang01 (approved):

PR #16, opened 2022-10-03
Update charm to not require a workload resource

mthaddon:
With the `assumes` entry in metadata.yaml we can ensure this charm isn't deployable on machine models even if it doesn't have a container.

This means we no longer need the placeholder workload container.

Has been tested to not be deployable on machine models (with juju > 2.9.23). Need to test a charm upgrade from existing to this version, but currently blocked on https://bugs.launchpad.net/juju/+bug/1991511.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ enable it with `juju expose kubernetes-worker`.
 
 ## Usage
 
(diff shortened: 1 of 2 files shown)


merkata (approved):
Looks great, approving.

jdkandersson (approved):

PR #18, opened 2022-10-04
Add a tox target for 'woke' to check for inclusive naming in the charm

mthaddon:
Add a tox target for 'woke' to check for inclusive naming in the charm.

Since this is provided by a snap, we're creating a shell command to check you have the snap installed or error out, and then run the woke command.

This isn't yet hooked up to workflows, as this charm doesn't have those yet.

### DIFF
--- a/.woke.yml
+++ b/.woke.yml
@@ -1,0 +1,7 @@
+rules:
+  # Ignore "whitelist" - upstream uses this terminology and it would be
+  # confusing to change in this charm while upstream continues to use it.
+  # Specifically the following annotations:
+  #   - nginx.ingress.kubernetes.io/limit-whitelist
+  #   - nginx.ingress.kubernetes.io/whitelist-source-range
+  - name: whitelist
--- a/.woke.yml
+++ b/.woke.yml
@@ -4,4 +4,6 @@ rules:
   # Specifically the following annotations:
   #   - nginx.ingress.kubernetes.io/limit-whitelist
   #   - nginx.ingress.kubernetes.io/whitelist-source-range
+  # See https://github.com/kubernetes/ingress-nginx/issues/7916 for upstream
+  # issue about this.
   - name: whitelist
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ enable it with `juju expose kubernetes-worker`.
 
 ## Usage
 
-You'll need version to be using Juju [version 2.9.0](https://discourse.charmhub.io/t/juju-2-9-0-release-notes/4525) or later, and Kubernetes version 1.19 or later.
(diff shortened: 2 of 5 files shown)


mthaddon:
Marking as draft as this isn't quite ready for review yet, as it includes changes that remove containers, which is blocked on an external issue.

README.md
```diff
@@ -15,7 +15,7 @@ enable it with `juju expose kubernetes-worker`.
 
 ## Usage
 
-You'll need version to be using Juju [version 2.9.0](https://discourse.charmhub.io/t/juju-2-9-0-release-notes/4525) or later, and Kubernetes version 1.19 or later.
+You'll need version to be using Juju [version 2.9.23](https://discourse.charmhub.io/t/roadmap-releases/5064) or later, and Kubernetes version 1.19 or later.
```
mthaddon, line 19: This change is unrelated, and has already landed in the repo.

metadata.yaml
```diff
@@ -7,14 +7,8 @@ summary: |
...
-containers:
-  placeholder:
-    resource: placeholder-image
-
-resources:
-  placeholder-image:
-    type: oci-image
-    description: Docker image for placeholder - won't be used
+assumes:
+  - k8s-api
```
mthaddon, line 12: This change is unrelated and has already landed in the repo

arturo-seijas (dismissed):
LGTM

.woke.yml
```diff
@@ -0,0 +1,9 @@
+rules:
```
jdkandersson, line 1: I suppose the tool doesn't look in `pyproject.toml`?

tox.ini
```diff
@@ -51,6 +52,11 @@ commands = flake8 src/ tests/ lib/
 deps =
     flake8==3.7.9
 
+[testenv:woke]
```
jdkandersson, line 55: Do you think this should be one of the checks in `lint` rather than a separate check? It seems to share many characteristics with a linting tool

jdkandersson (changes requested):

.woke.yml
mthaddon: I don't think so, it's not a python-specific project.

tox.ini
mthaddon: I agree it shares some characteristics, but it feels different enough to me that it warrants a different target. Happy to be overridden if I'm in the minority on that.

arturo-seijas: Maybe it will be appropriate to add this target the `envlist`, also, `lint`is missing

jdkandersson (dismissed):

arturo-seijas (approved):
LGTM

jdkandersson (approved):

PR #19, opened 2022-10-21
Set up github workflows

mthaddon:
Pull in github workflows for initial CI setup.

For now this isn't running the `woke` target, because we'll need to install the `woke` snap for that to work, which will require some changes to operator-workflows. This also doesn't yet include static code analysis with bandit, that will come in a later PR.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,7 +1,7 @@
 [tox]
 allowlist_externals = ./check_woke.sh
 skipsdist = True
-envlist = lint, unit, integration
+envlist = lint, unit
 skip_missing_interpreters = True
 
 [vars]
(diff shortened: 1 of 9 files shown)


tox.ini
```diff
@@ -1,7 +1,7 @@
 [tox]
-allowlist_externals=./check_woke.sh
-skipsdist=True
-envlist = lint, woke, unit, functional
+allowlist_externals = ./check_woke.sh
+skipsdist = True
+envlist = lint, unit, integration
```
arturo-seijas, line 4: The `integration` target will be run also in the `integration_test`workflow. The targets defined here are run as part of the unit test step. Hence, I recommend you remove the `integration`target from this list

mthaddon: Ah thanks, will update.

arturo-seijas (approved):
LGTM

gregory-schiano (approved):
LGTM

PR #20, opened 2022-10-21
Fix tests for ops > 1.5.2

mthaddon:
Operator Framework > 1.5.2 is stricter with config values, update so that tests pass.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,7 +1,7 @@
 [tox]
 allowlist_externals = ./check_woke.sh
 skipsdist = True
-envlist = lint, unit
+envlist = lint, unit, static, coverage-report
 skip_missing_interpreters = True
 
 [vars]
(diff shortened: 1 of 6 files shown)


arturo-seijas (dismissed):
LGTM

tox.ini
```diff
@@ -52,6 +52,15 @@ commands = flake8 src/ tests/ lib/
 deps =
     flake8==3.7.9
 
+[testenv:static]
```
arturo-seijas, line 64: Could you add this to envlist?

arturo-seijas (changes requested):

tox.ini
mthaddon: I have done now, yep.

arturo-seijas (dismissed):
LGTM

arturo-seijas (approved):
LGTM

gregory-schiano (approved):
LGTM
As discussed with mthaddon, there "Comment on PR" doesn't work, but it's an issue on workflow library

PR #21, opened 2022-10-21
Add our check for inclusive naming to the lint target so it will happen on every CI run

mthaddon:
Run our inclusive naming checks as part of the `lint` target. This will mean it's triggered on every CI run.


gregory-schiano (approved):
LGTM

PR #23, opened 2022-10-24
Add missing lines to test coverage and fail under specified percentage

mthaddon:
Update pyproject.toml so that we're including missing lines in test coverage output and failing if we're under a specified percentage - pin to the current test coverage, which is 93%.


arturo-seijas (approved):
LGTM

arturo-seijas (approved):
LGTM

gtrkiller (approved):
LGTM

PR #29, opened 2022-11-09
Add ingress IP to juju status

gtrkiller:
With this PR we want to be able to obtain the ingress IPs from the load balancers and display them in the message of the application alongside the existing service IPs.

We also update from 20.04 to 22.04 due to an unrelated issue during the course of the PR, which has now been resolved (but updating to 22.04 is still worthwhile).


arturo-seijas (dismissed):
LGTM

src/charm.py
```diff
@@ -418,6 +416,30 @@ def _report_service_ips(self):
...
+    def _report_ingress_ips(self) -> str:
+        """Report on ingress IP(s)."""
+        self.k8s_auth()
+        api = _networking_v1_api()
+
+        # Wait up to `interval * count` seconds for ingress IPs.
+        count, interval = 100, 1
+        for _ in range(count):
+            ingresses = api.list_namespaced_ingress(namespace=self._namespace)
+            # XXX: Will this return multiple IPs if there are any?
```
mthaddon, line 428: We shouldn't have an XXX comment here - can we confirm this either way and update as appropriate?

src/charm.py
```diff
@@ -418,6 +416,30 @@ def _report_service_ips(self):
...
+                ips = [x.status.load_balancer.ingress[0].ip for x in ingresses.items]
+            except TypeError:
+                # We have no IPs yet.
+                ips = []
+            if ips:
+                break
+            LOGGER.info("Sleeping for %s seconds to wait for ingress IP", interval)
+            time.sleep(interval)
+        if ips:
+            return "Ingress IP(s): {}, ".format(str(ips))
```
mthaddon, line 439: This part should move to building the message, and we should use a join, similar to how we currently do for service IPs.

mthaddon (changes requested):

src/charm.py
```diff
@@ -418,6 +419,26 @@ def _report_service_ips(self):
             x.spec.cluster_ip for x in services.items if x.metadata.name in all_k8s_service_names
         ]
 
+    def _report_ingress_ips(self) -> list:
```
jdkandersson, line 422: Do you think it is worth specifying the type of the list items?

src/charm.py
```diff
@@ -418,6 +419,26 @@ def _report_service_ips(self):
...
+        """Report on ingress IP(s)."""
```
jdkandersson, line 423: We have a standard for docstrings that includes things like describing the return value, do you think it applies here? https://github.com/canonical/is-charms-contributing-guide/blob/main/CONTRIBUTING.md#docstrings

src/charm.py
```diff
@@ -696,11 +717,15 @@ def _on_config_changed(self, _):
             try:
                 self._define_services()
                 self._define_ingresses()
-
+                ingress_ip = (
+                    "Ingress IP(s): {}, ".format(", ".join(self._report_ingress_ips()))
```
jdkandersson, line 721: would an f-string work here?

src/charm.py
```diff
@@ -696,11 +717,15 @@ def _on_config_changed(self, _):
...
+                    if self._report_ingress_ips() != ""
+                    else ""
+                )
                 # It's not recommended to do this via ActiveStatus, but we don't
                 # have another way of reporting status yet.
-                msg = "Ingress with service IP(s): {}".format(
-                    ", ".join(self._report_service_ips())
+                msg = "{}Service IP(s): {}".format(
```
jdkandersson, line 727: Would an f-string work here?

tests/integration/test_charm.py
```diff
@@ -29,11 +29,11 @@ async def test_reachable(app: Application):
     """
     # Get the IP address which is in the status message
     ip_regex = r"[0-9]+(?:\.[0-9]+){3}"
-    ip_address_match = re.search(ip_regex, app.units[0].workload_status_message)
+    ip_address_match = re.findall(ip_regex, app.units[0].workload_status_message)
```
jdkandersson, line 32: Out of curiosity, why the switch from `search` to `findall`?

tests/unit/test_charm.py
```diff
@@ -20,14 +20,18 @@ def setUp(self):
...
-    def test_config_changed(self, _define_service, _define_ingress, _report_service_ips):
+    def test_config_changed(
+        self, _define_service, _define_ingress, _report_service_ips, _report_ingress_ips
+    ):
         """Test our config changed handler."""
         # First of all test, with leader set to True.
         self.harness.set_leader(True)
-        _report_service_ips.return_value = ["10.0.1.12"]
+        _report_ingress_ips.return_value = ["10.0.1.12"]
+        _report_service_ips.return_value = ["10.0.1.13"]
```
jdkandersson, line 34: Is this second line needed? It seems that the `return_value` is updated twice in a row without using it in between?

tests/unit/test_charm.py
```diff
@@ -49,7 +53,7 @@ def test_config_changed(self, _define_service, _define_ingress, _report_service_
         # Confirm status is as expected.
         self.assertEqual(
             self.harness.charm.unit.status,
-            ActiveStatus("Ingress with service IP(s): 10.0.1.12"),
+            ActiveStatus("Ingress IP(s): 10.0.1.12, Service IP(s): 10.0.1.13"),
```
jdkandersson, line 56: Tests that check for specific strings on informational messages like this can be flaky and require duplicating the message in the source and test code. A potential alternative is to look for key words, word sequences or values in the string. For example, this assert could look for that the message contains the required IP addresses

tests/unit/test_charm.py
```diff
@@ -1406,12 +1420,70 @@ def test_ingresses_for_multiple_relations_different_hostnames(
...
+    def test_report_ingress_ips(
+        self,
+        mock_api,
+        mock_define_service,
+        mock_remove_service,
+        mock_define_ingress,
+        mock_k8s_auth,
+        mock_service_ips,
+    ):
+        """Test for checking the API call to obtain ingress IPs."""
```
jdkandersson, line 1438: We have a standard for writing the docstring for tests, do you think it applies here? https://github.com/canonical/is-charms-contributing-guide/blob/main/CONTRIBUTING.md#test-structure

Also, consider updating the docstring of any tests being changed here

tests/unit/test_charm.py
```diff
@@ -1406,12 +1420,70 @@ def test_ingresses_for_multiple_relations_different_hostnames(
...
+        self,
+        mock_api,
+        mock_define_service,
+        mock_remove_service,
+        mock_define_ingress,
+        mock_k8s_auth,
+        mock_service_ips,
+    ):
+        """
+        Test for checking the API call to obtain ingress IPs
```
jdkandersson, line 1467: We have a standard for writing the docstring for tests, do you think it applies here? https://github.com/canonical/is-charms-contributing-guide/blob/main/CONTRIBUTING.md#test-structure

tests/unit/test_charm.py
```diff
@@ -1570,20 +1645,22 @@ def test_ingresses_for_multiple_relations_blocked(
         rel_data["path-routes"] = "/funicorn"
         self.harness.update_relation_data(rel_id, "funicorn", rel_data)
 
-        expected_status = ActiveStatus("Ingress with service IP(s): 10.0.1.12")
+        expected_status = ActiveStatus("Service IP(s): 10.0.1.12")
```
jdkandersson, line 1648: As above, consider just checking for key values or specific words rather than the entire service message

tests/integration/test_charm.py
gtrkiller: because now the message has 2 IPs, and we want only the second one. Re.search() always returns the first ocurrence of the pattern.

tests/unit/test_charm.py
gtrkiller: Yes, it is needed for the AssertEqual you can find at the end of the test with "confirm the status is as expected" as a comment.

tests/unit/test_charm.py
mthaddon: I don't think they'd be any more flaky than checking for specific parts of the message. It does mean you might need to update the tests if you update the message, but I don't think that's a bad thing. Personally I prefer seeing more verbose messages in tests because it's easier to understand what you're testing for, and confirms the wording is correct, which can sometimes be important. If you're just looking for a specific IP address here how do you know if it's the ingress address or the service address?

arturo-seijas (dismissed):
LGTM

merkata (approved):
Awesome!

arturo-seijas (dismissed):
LGTM

tests/unit/test_charm.py
mthaddon: Also, the return value is updated for two different things - once for ingress_ips and once for service_ips :)

src/charm.py
```diff
@@ -696,11 +717,15 @@ def _on_config_changed(self, _):
...
+                    f"Ingress IP(s): {', '.join(self._report_ingress_ips())}, "
+                    if self._report_ingress_ips()
+                    else ""
+                )
                 # It's not recommended to do this via ActiveStatus, but we don't
                 # have another way of reporting status yet.
-                msg = "Ingress with service IP(s): {}".format(
-                    ", ".join(self._report_service_ips())
+                msg = f"{ingress_ip}Service IP(s): {', '.join(self._report_service_ips())}".format(
+                    ingress_ip,
```
mthaddon, line 728: This seems a little awkward as you've got a variable with a trailing space in it which means you have to concatenate without a space here (in case `ingress_ip` is empty. (Also, you're mixing f-strings and string formatting). Also, I think you're calling `self._report_ingress_ips()` twice in the block above? How about something like this:
```
msgs = []
ingress_ips = self._report_ingress_ips()
if ingress_ips:
    msgs.append(f"Ingress IP(s): {', '.join(ingress_ips)}
msgs.append(f"Service IP(s): {', '.join(self._report_service_ips())}")
msg = ", ".join(msgs)
```

mthaddon (approved):
LGTM, assuming integration tests pass - thanks!

merkata (approved):

PR #30, opened 2022-11-10
Pin workflows to main

arturo-seijas:
Removes also woke execution from tox since the list of term will be different


weiiwang01 (dismissed):

amandahla (dismissed):
LGTM

mthaddon (approved):
Ok, this does mean we can't run woke locally, but I don't think there's a good way around that given that we're using the GH action.

gtrkiller (approved):
LGTM

PR #33, opened 2022-11-14
Expand integration tests

gtrkiller:
This PR adds integration tests to reflect the changes we've made to the charm recently.

### DIFF
--- a/tests/integration/test_charm.py
+++ b/tests/integration/test_charm.py
@@ -2,6 +2,7 @@
 # See LICENSE file for licensing details.
 
 import re
+import subprocess
 import time
 
 import pytest
@@ -22,10 +23,10 @@ async def test_active(app: Application):
 
 @pytest.mark.asyncio
 @pytest.mark.abort_on_fail
-async def test_reachable(app: Application):
+async def test_service_reachable(app: Application):
     """
     arrange: given charm has been built, deployed and related to a dependent application
-    act: when the dependent application is queried via the ingress
+    act: when the dependent application is queried via the service
     assert: then the response is HTTP 200 OK.
     """
     # Get the IP address which is in the status message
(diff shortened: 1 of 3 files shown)


tests/integration/test_charm.py
```diff
@@ -41,3 +42,27 @@ async def test_reachable(app: Application):
...
+    """
+    # Get the IP address which is in the status message
+    ip_regex = r"[0-9]+(?:\.[0-9]+){3}"
+    ip_address_match = re.findall(ip_regex, app.units[0].workload_status_message)
+    assert (
+        ip_address_match
+    ), f"could not find IP address in status message: {app.units[0].workload_status_message}"
+    ip_address = ip_address_match[0]
+    port = "80"
+    ps = subprocess.Popen(['echo', f'{ip_address} hello-kubecon'], stdout=subprocess.PIPE)  # nosec
```
arturo-seijas, line 63: Does this pass the linter? I would expect this to be enquoted in double quotes

gtrkiller: yes, it passes the linter and executes the task correctly

tests/integration/test_charm.py
```diff
@@ -41,3 +42,27 @@ async def test_reachable(app: Application):
...
+    subprocess.run(['sudo', 'tee', '-a', '/etc/hosts'], stdin=ps.stdout)  # nosec
+
+    response = requests.get(f"http://hello-kubecon:{port}")
```
mthaddon, line 66: We're setting the port to 80 here, which is the default for http. Do we really need to do that? The instructions we include in the `README` don't instruct people to connect to `http://hello-kubecon:80` but to `http://hello-kubecon` so I think we should just do that here too.

tests/integration/test_charm.py
gtrkiller: Yes, but when changing it to the linter we use in indico, I get a lot of errors that are out of the scope of this PR. I'd suggest to open a new PR with those changes and address them there.

arturo-seijas: Ok. Can you change those quotes in the interim? I'd be better to keep it as compliant with the other linter as possible

gtrkiller: Sure, no problem

arturo-seijas (approved):
LGTM

tests/integration/test_charm.py
```diff
@@ -41,3 +42,26 @@ async def test_reachable(app: Application):
...
+    """
+    # Get the IP address which is in the status message
+    ip_regex = r"[0-9]+(?:\.[0-9]+){3}"
+    ip_address_match = re.findall(ip_regex, app.units[0].workload_status_message)
+    assert (
+        ip_address_match
+    ), f"could not find IP address in status message: {app.units[0].workload_status_message}"
+    ip_address = ip_address_match[0]
+    ps = subprocess.Popen(["echo", f"{ip_address} hello-kubecon"], stdout=subprocess.PIPE)  # nosec
+    subprocess.run(["sudo", "tee", "-a", "/etc/hosts"], stdin=ps.stdout)  # nosec
```
amandahla, line 63: Minor thing: Would be possible to call passing the host header instead of changing /etc/hosts like Indico?
https://github.com/canonical/indico-operator/blob/main/tests/integration/test_charm.py#L35

amandahla (approved):

tests/integration/test_charm.py
gtrkiller: for ingress specifically, it is necessary to change the /etc/hosts file to make the IP reachable and display the page correctly.

PR #36, opened 2022-11-16
Revert to focal as k8s charm series upgrades aren't well handled currently

mthaddon:
There's currently no good way to have a different series of a k8s charm in edge and stable, and be able to test upgrades from one to another (see https://bugs.launchpad.net/juju/+bug/1996590). So let's revert to focal for now. We bumped it up to jammy due to build failures on focal that have since been resolved.

### DIFF
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -2,7 +2,7 @@ type: charm
 bases:
   - build-on:
     - name: "ubuntu"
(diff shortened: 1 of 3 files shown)


arturo-seijas (approved):
LGTM

weiiwang01 (approved):

PR #37, opened 2022-11-17
Add copyright info to library, and increment version number

mthaddon:
We need to increment the version number so this can be published, and charmhub has a limitation on the minor version number which means we're forced to increment `LIBAPI` (it only allows `LIBPATCH` to go up to 10)

### DIFF
--- a/lib/charms/nginx_ingress_integrator/v0/ingress.py
+++ b/lib/charms/nginx_ingress_integrator/v0/ingress.py
@@ -1,5 +1,5 @@
 # Copyright 2022 Canonical Ltd.
-# Licensed under the GPLv3, see LICENCE file for details.
+# Licensed under the Apache2.0, see LICENCE file in charm source for details.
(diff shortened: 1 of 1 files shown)


lib/charms/nginx_ingress_integrator/v0/ingress.py
```diff
@@ -1,3 +1,5 @@
+# Copyright 2022 Canonical Ltd.
+# Licensed under the GPLv3, see LICENCE file for details.
```
arturo-seijas, line 2: GPLv3 or Apache2? We're licensing charms using this library as Apache 2

mthaddon: Er, absolultely. Bad copy/paste...

arturo-seijas (approved):
LGTM

weiiwang01 (approved):

PR #39, opened 2022-11-17
Add owasp-modsecurity-custom-rules config

amandahla:
If owasp-modsecurity-crs is set to True, you can add custom rules via owasp-modsecurity-custom-rules.

Command example:
```bash
juju config nginx-ingress-integrator owasp-modsecurity-custom-rules='SecAction "id:900130,phase:1,nolog,pass,t:none,setvar:tx.crs_exclusions_wordpress=1"\n'
```
This PR adds:
- owasp-modsecurity-custom-rules new config
- unit test
- integration test

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -49,12 +49,18 @@ options:
     type: int
   owasp-modsecurity-crs:
     default: false
-    description: >
+    description: |
       Enable OWASP ModSecurity Core Rule Set (CRS). A set of generic attack detection rules for use
       with ModSecurity or compatible web application firewalls. The CRS aims to protect web
       applications from a wide range of attacks, including the OWASP Top Ten, with a minimum of
       false alerts. See https://github.com/coreruleset/coreruleset for more details.
     type: boolean
+  owasp-modsecurity-custom-rules:
+    default: ""
+    description: >
+      /n separated list of custom rules to be added to modsecurity-snippet annotation.
+      Example: "SecAction id:900130,phase:1,nolog,pass,t:none,setvar:tx.crs_exclusions_wordpress=1\n".
+    type: string
   path-routes:
     default: ""
     description: >
--- a/config.yaml
+++ b/config.yaml
@@ -8,7 +8,7 @@ options:
     type: string
   ingress-class:
     default: ""
-    description: >
+    description: |
       The ingress class to target for this ingress resource.
 
       If your cluster has multiple ingress controllers, this allows
@@ -31,7 +31,7 @@ options:
     type: string
   limit-rps:
     default: 0
-    description: >
+    description: |
       Number of requests accepted from a given IP each second. The burst limit
       is set to this limit multiplied by 5. When clients exceed this limit a
       503 error will be returned.
@@ -39,7 +39,7 @@ options:
     type: int
   limit-whitelist:
     default: ""
-    description: >
+    description: |
       If rate-limiting is set, client IP source ranges to be excluded. The value
       is a comma-separated list of CIDRs.
     type: string
@@ -57,19 +57,19 @@ options:
     type: boolean
   owasp-modsecurity-custom-rules:
     default: ""
-    description: >
-      /n separated list of custom rules to be added to modsecurity-snippet annotation.
+    description: |
+      New line ('\n') separated list of custom rules to be added to modsecurity-snippet annotation.
       Example: "SecAction id:900130,phase:1,nolog,pass,t:none,setvar:tx.crs_exclusions_wordpress=1\n".
     type: string
   path-routes:
     default: ""
-    description: >
+    description: |
       Comma separated list of the routes under the hostname that you wish to map to the relation.
       Example: "/admin,/portal" will map example.test/admin and example.test/portal only.
     type: string
   retry-errors:
     default: ""
-    description: >
+    description: |
       Specifies in which cases a request should be retried against the next server.
       Comma-separated list, e.g. "error,timeout,http_502,http_503,http_504".
       See http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_next_upstream
--- a/lib/charms/nginx_ingress_integrator/v0/ingress.py
+++ b/lib/charms/nginx_ingress_integrator/v0/ingress.py
@@ -69,7 +69,7 @@ LIBAPI = 1
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 1
+LIBPATCH = 2
 
 LOGGER = logging.getLogger(__name__)
 
@@ -214,7 +214,8 @@ class IngressProvides(Object):
     def _on_relation_changed(self, event):
         """Handle a change to the ingress relation.
 
-        Confirm we have the fields we expect to receive."""
+        Confirm we have the fields we expect to receive.
+        """
         # `self.unit` isn't available here, so use `self.model.unit`.
         if not self.model.unit.is_leader():
             return
--- a/lib/charms/nginx_ingress_integrator/v0/ingress.py
+++ b/lib/charms/nginx_ingress_integrator/v0/ingress.py
@@ -18,6 +18,7 @@ Import `IngressRequires` in your charm, with two required options:
 - limit-whitelist
 - max-body-size
 - owasp-modsecurity-crs
+- owasp-modsecurity-custom-rules
 - path-routes
 - retry-errors
 - rewrite-enabled
@@ -80,6 +81,7 @@ OPTIONAL_INGRESS_RELATION_FIELDS = {
     "limit-whitelist",
     "max-body-size",
     "owasp-modsecurity-crs",
+    "owasp-modsecurity-custom-rules",
     "path-routes",
     "retry-errors",
     "rewrite-target",
--- a/src/charm.py
+++ b/src/charm.py
@@ -158,8 +158,19 @@ class _ConfigOrRelation(object):
         return self._get_config_or_relation_data("owasp-modsecurity-crs", False)
 
     @property
+    def _owasp_modsecurity_custom_rules(self):
+        """Return the owasp-modsecurity-custom-rules value from config or relation."""
+        custom_rules = self._get_config_or_relation_data("owasp-modsecurity-custom-rules", "")
+        # Since expected value is something like the example below,
+        # we replace the escaped new line \\n for \n
+        # SecAction "id:900130,phase:1,nolog,pass,t:none,setvar:tx.crs_exclusions_wordpress=1"\n
+        custom_rules.replace("\\n", "\n")
+
+        return custom_rules
+
+    @property
     def _rewrite_enabled(self):
-        """Return whether rewriting should be enabled from config or relation"""
+        """Return whether rewriting should be enabled from config or relation."""
         value = self._get_config_or_relation_data("rewrite-enabled", True)
         # config data is typed, relation data is a string
         # Convert to string, then compare to a known value.
@@ -309,9 +320,13 @@ class _ConfigOrRelation(object):
         if self._owasp_modsecurity_crs:
             annotations["nginx.ingress.kubernetes.io/enable-modsecurity"] = "true"
             annotations["nginx.ingress.kubernetes.io/enable-owasp-modsecurity-crs"] = "true"
+            nginx_modsec_file = "/etc/nginx/owasp-modsecurity-crs/nginx-modsecurity.conf"
             annotations[
                 "nginx.ingress.kubernetes.io/modsecurity-snippet"
-            ] = "SecRuleEngine On\nInclude /etc/nginx/owasp-modsecurity-crs/nginx-modsecurity.conf"
+            ] = "SecRuleEngine On\n{}Include {}".format(
+                self._owasp_modsecurity_custom_rules, nginx_modsec_file
+            )
+
         if self._retry_errors:
             annotations["nginx.ingress.kubernetes.io/proxy-next-upstream"] = self._retry_errors
         if self._rewrite_enabled:
@@ -637,7 +652,6 @@ class NginxIngressCharm(CharmBase):
 
     def _create_k8s_ingress_obj(self, svc_hostname, initial_ingress, paths):
         """Creates a Kubernetes Ingress Resources with the given data."""
-
         # Create a Ingress Object with the new ingress rules and return it.
         rule = kubernetes.client.V1IngressRule(
             host=svc_hostname,
--- a/src/charm.py
+++ b/src/charm.py
@@ -118,7 +118,7 @@ class _ConfigOrRelation(object):
         # Avoid collision with service name created by Juju. Currently
         # Juju creates a K8s service listening on port 65535/TCP so we
         # need to create a separate one.
-        return "{}-service".format(self._service_name)
+        return f"{self._service_name}-service"
 
     @property
     def _ingress_name(self):
@@ -130,7 +130,7 @@ class _ConfigOrRelation(object):
         # based on their requested service-hostname.
         svc_hostname = self._get_config_or_relation_data("service-hostname", "")
         ingress_name = _INGRESS_SUB_REGEX.sub("-", svc_hostname)
-        return "{}-ingress".format(ingress_name)
+        return f"{ingress_name}-ingress"
 
     @property
     def _limit_rps(self):
@@ -150,7 +150,7 @@ class _ConfigOrRelation(object):
     def _max_body_size(self):
         """Return the max-body-size to use for k8s ingress."""
         max_body_size = self._get_config_or_relation_data("max-body-size", 0)
-        return "{}m".format(max_body_size)
+        return f"{max_body_size}m"
 
     @property
     def _owasp_modsecurity_crs(self):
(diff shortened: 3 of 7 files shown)


gtrkiller:
you may want to check the linting again :)

amandahla:
> you may want to check the linting again :)

I don't know why when I ran locally this didn't appear :thinking:

image

config.yaml
```diff
@@ -55,6 +55,12 @@ options:
       applications from a wide range of attacks, including the OWASP Top Ten, with a minimum of
       false alerts. See https://github.com/coreruleset/coreruleset for more details.
     type: boolean
+  owasp-modsecurity-custom-rules:
+    default: ""
+    description: >
+      \n separated list of custom rules to be added to modsecurity-snippet annotation.
```
gtrkiller, line 61: why is there an \n at first in the description?

tests/integration/test_charm.py
```diff
@@ -65,3 +65,29 @@ async def test_ingress_reachable(app: Application):
...
+
+
+@pytest.mark.asyncio
+@pytest.mark.abort_on_fail
+async def test_owasp_modsecurity_crs(app: Application):
+    """
+    arrange: given charm has been built, deployed, related to a dependent application
+    and owasp-modsecurity-crs is set to True
+    act: when the dependent application is queried via the ingress with malicious request
+    assert: then the response is HTTP 403 Forbidden.
```
gtrkiller, line 77: I think this line should contemplate the 404 case as well, does it make sense?

config.yaml
```diff
@@ -55,6 +55,12 @@ options:
       applications from a wide range of attacks, including the OWASP Top Ten, with a minimum of
       false alerts. See https://github.com/coreruleset/coreruleset for more details.
     type: boolean
+  owasp-modsecurity-custom-rules:
+    default: ""
+    description: >
```
jdkandersson, line 60: In yaml we usually use `|` to enable longer descriptions, e.g., 

```yaml
owasp-modsecurity-custom-rules:
  default: ""
  description: |
    separated list of custom rules to be added to modsecurity-snippet annotation.
    Example: "SecAction id:900130,phase:1,nolog,pass,t:none,setvar:tx.crs_exclusions_wordpress=1\n".
```

config.yaml
jdkandersson: Also looks like there could be trailing whitespace?

src/charm.py
```diff
@@ -309,9 +320,15 @@ def _get_k8s_ingress(self):
         if self._owasp_modsecurity_crs:
             annotations["nginx.ingress.kubernetes.io/enable-modsecurity"] = "true"
             annotations["nginx.ingress.kubernetes.io/enable-owasp-modsecurity-crs"] = "true"
+
+            nginx_modsec_file = "/etc/nginx/owasp-modsecurity-crs/nginx-modsecurity.conf"
+            logging.info(repr(self._owasp_modsecurity_custom_rules))
```
jdkandersson, line 325: Should this log message include some information about what is being logged in addition the rules?

src/charm.py
```diff
@@ -309,9 +320,15 @@ def _get_k8s_ingress(self):
...
             annotations[
                 "nginx.ingress.kubernetes.io/modsecurity-snippet"
-            ] = "SecRuleEngine On\nInclude /etc/nginx/owasp-modsecurity-crs/nginx-modsecurity.conf"
+            ] = "SecRuleEngine On\n{}Include {}".format(
```
jdkandersson, line 328: Could this be an f-string?

tests/integration/test_charm.py
```diff
@@ -65,3 +65,29 @@ async def test_ingress_reachable(app: Application):
...
 
     assert response.status_code == 200
+
+
+@pytest.mark.asyncio
+@pytest.mark.abort_on_fail
+async def test_owasp_modsecurity_crs(app: Application):
+    """
+    arrange: given charm has been built, deployed, related to a dependent application
+    and owasp-modsecurity-crs is set to True
```
jdkandersson, line 75: Should this line be indented 4 more spaces as described here: https://github.com/canonical/is-charms-contributing-guide/blob/main/CONTRIBUTING.md#test-structure?

tests/integration/conftest.py
```diff
@@ -49,4 +49,12 @@ async def app(ops_test: OpsTest, app_name: str):
     await ops_test.model.add_relation(hello_kubecon_app_name, app_name)
     await ops_test.model.wait_for_idle(status=ActiveStatus.name, idle_period=120)
 
+    # Add ignore rule
```
jdkandersson, line 52: Should this be done for all tests? Or just a test specifically for the ignore rules?

tests/integration/test_charm.py
```diff
@@ -65,3 +65,29 @@ async def test_ingress_reachable(app: Application):
...
+    """
+    # Get the IP address which is in the status message
+    ip_regex = r"[0-9]+(?:\.[0-9]+){3}"
+    ip_address_match = re.findall(ip_regex, app.units[0].workload_status_message)
+    assert (
+        ip_address_match
+    ), f"could not find IP address in status message: {app.units[0].workload_status_message}"
+    ip_address = ip_address_match[0]
+    ps = subprocess.Popen(["echo", f"{ip_address} hello-kubecon"], stdout=subprocess.PIPE)  # nosec
+    subprocess.run(["sudo", "tee", "-a", "/etc/hosts"], stdin=ps.stdout)  # nosec
```
jdkandersson, line 87: Out of curiosity, does the `stdin` have to be the `stdout` from calling another process? Or could it just be the string `f"{ip_address} hello-kubecon"`?

tests/unit/test_charm.py
```diff
@@ -248,6 +248,7 @@ def test_owasp_modsecurity_crs(self):
         # Test undefined.
         conf_or_rel = self.harness.charm._all_config_or_relations[0]
         self.assertEqual(conf_or_rel._owasp_modsecurity_crs, False)
+        self.assertEqual(conf_or_rel._owasp_modsecurity_custom_rules, "")
```
jdkandersson, line 251: Should the test description be updated to include information about that a new check has been included?

src/charm.py
```diff
@@ -157,6 +157,17 @@ def _owasp_modsecurity_crs(self):
...
         return self._get_config_or_relation_data("owasp-modsecurity-crs", False)
 
+    @property
+    def _owasp_modsecurity_custom_rules(self):
+        """Return the owasp-modsecurity-custom-rules value from config or relation"""
+        custom_rules = self._get_config_or_relation_data("owasp-modsecurity-custom-rules", "")
+        # Since expected value is something like the example below,
+        # we replace the escaped new line \\n for \n
+        # SecAction "id:900130,phase:1,nolog,pass,t:none,setvar:tx.crs_exclusions_wordpress=1"\n
+        custom_rules.replace("\\n", "\n")
```
jdkandersson, line 167: I could have missed it, I did not spot any tests that checked for this?

weiiwang01:
I personally prefer the option to be `owasp-modsecurity-snippet` and value of the option will replace the entire content of the `nginx.ingress.kubernetes.io/modsecurity-snippet` annotation. It will be a simpler implementation. User will have a better idea of what this option will do. And it will allow more customization. But it's all up to you.

config.yaml
amandahla: Sorry, when I wrote made sense in my head. It's a list separated by "\n" like
```'SecAction "id:900130,phase:1,nolog,pass,t:none,setvar:tx.crs_exclusions_wordpress=1"\nSecAction "id:900130,phase:1,nolog,pass,t:none,setvar:tx.crs_exclusions_drupal=1"\n'```

Comment (amandahla) [reply config.yaml:61]: Trailing whitespace removed.

File: tests/integration/test_charm.py (hunk 2 above)
Comment (amandahla) [reply tests/integration/test_charm.py:77]: Sure! Done.
```python
assert: then the response is HTTP 403 Forbidden for any request except for /ignore
        whose response is HTTP 404 not found
```

src/charm.py
amandahla: My bad, this was not supposed to be there.

src/charm.py
amandahla: I suppose that would not be a problem since f-string prints '\n' correctly. I used '.format' to be consistent with the previous code. Do you think I should change the other format occurrences for f-string?

config.yaml
amandahla: Do you think would be better to use another separator or just change the description to make it clear that '\n' is the separator? Using \n is easier because it's not necessary to format the config and I think would be safe to use with [SecRule/SecAction configs](https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual-(v2.x)#secrule)

amandahla:
> I personally prefer the option to be `owasp-modsecurity-snippet` and value of the option will replace the entire content of the `nginx.ingress.kubernetes.io/modsecurity-snippet` annotation. It will be a simpler implementation. User will have a better idea of what this option will do. And it will allow more customization. But it's all up to you.

Good point but in that case, you could set owasp_modsecurity_crs to true and the owasp-modsecurity-snippet would have the default value "SecRuleEngine On\nInclude /etc/nginx/owasp-modsecurity-crs/nginx-modsecurity.conf", for example.
But if someone changes to "SecRuleEngine Off" (or anything else), _owasp_modsecurity_crs could still be true even with modsecurity disabled. It would be necessary to validate that.
What do you think?

tests/integration/test_charm.py
amandahla: I'll do a test with the "input" parameter.
https://docs.python.org/3/library/subprocess.html#subprocess.run

tests/unit/test_charm.py
amandahla: Changed for "Test the owasp-modsecurity-crs and owasp-modsecurity-custom-rules properties."

tests/integration/conftest.py
amandahla: Changed all the integration tests, will push new version now.

src/charm.py
amandahla: Good catch. I'll do this one on Monday.

config.yaml
jdkandersson: It is probably worth it yes 👍

src/charm.py
jdkandersson: Our guidance for the scenario is that we should always fix these issues for any code being edited updated, updating the rest of the code would be valuable but is up to the individual (e.g., it might increase the scope of the PR too much so we delay it): https://github.com/canonical/is-charms-contributing-guide/blob/main/CONTRIBUTING.md#non-compliant-code

tests/integration/conftest.py
```diff
@@ -50,3 +53,44 @@ async def app(ops_test: OpsTest, app_name: str):
...
 
     yield application
+
+
+@pytest_asyncio.fixture(scope="module")
+async def app_ip(app: Application):
+    """Get unit IP address and add to /etc/hosts."""
+    # Get the IP address which is in the status message
+    ip_regex = r"[0-9]+(?:\.[0-9]+){3}"
+    time.sleep(100)  # wait for ingress
```
jdkandersson, line 63: I wonder if `ops_test.model.wait_for_idle()` works here? It doesn't always, e.g. in the case of Jenkins as you might have seen. Otherwise:
1. Since this is an async function, would `await asyncio.sleep()` work it? It would release the event loop to potentially do other things
2. Do you think it is worth having some sort of check for an early exit? E.g., maybe a function that looks for whether the IP address has appeared in the status message and then stops the waiting? It would potentially reduce the time the tests take to execute

config.yaml
jdkandersson: Oh I see now, maybe quote or describe it `new line ('\n') separated ...`?

tests/integration/test_charm.py
jdkandersson: That does look better 👍

jdkandersson:
I wonder why it looks like there is only 1 commit? Maybe it was the force push? Having multiple commits helps with making reviews easier on GitHub since reviewers can just see what has changed since their last review rather than reading the entire change again. We use squash merging anyway which means that in `main` it will end up looking like a single commit anyway

weiiwang01:
> > I personally prefer the option to be `owasp-modsecurity-snippet` and value of the option will replace the entire content of the `nginx.ingress.kubernetes.io/modsecurity-snippet` annotation. It will be a simpler implementation. User will have a better idea of what this option will do. And it will allow more customization. But it's all up to you.
> 
> Good point but in that case, you could set owasp_modsecurity_crs to true and the owasp-modsecurity-snippet would have the default value "SecRuleEngine On\nInclude /etc/nginx/owasp-modsecurity-crs/nginx-modsecurity.conf", for example. But if someone changes to "SecRuleEngine Off" (or anything else), _owasp_modsecurity_crs could still be true even with modsecurity disabled. It would be necessary to validate that. What do you think?

Yes, that's possible. But I think whoever sets this option should have a basic grasp of the modsecurity setting.

src/charm.py
amandahla: Fixed all, it was just about 11 strings using .format.

amandahla:
> I wonder why it looks like there is only 1 commit? Maybe it was the force push? Having multiple commits helps with making reviews easier on GitHub since reviewers can just see what has changed since their last review rather than reading the entire change again. We use squash merging anyway which means that in `main` it will end up looking like a single commit anyway

Something I do out of habit. The next ones will keep commits.

amandahla:
> > > I personally prefer the option to be `owasp-modsecurity-snippet` and value of the option will replace the entire content of the `nginx.ingress.kubernetes.io/modsecurity-snippet` annotation. It will be a simpler implementation. User will have a better idea of what this option will do. And it will allow more customization. But it's all up to you.
> > 
> > 
> > Good point but in that case, you could set owasp_modsecurity_crs to true and the owasp-modsecurity-snippet would have the default value "SecRuleEngine On\nInclude /etc/nginx/owasp-modsecurity-crs/nginx-modsecurity.conf", for example. But if someone changes to "SecRuleEngine Off" (or anything else), _owasp_modsecurity_crs could still be true even with modsecurity disabled. It would be necessary to validate that. What do you think?
> 
> Yes, that's possible. But I think whoever sets this option should have a basic grasp of the modsecurity setting.

Maybe we can keep with that separate option for now that has no impact on owasp_modsecurity_crs and consider changing everything for just one option in the future.

tests/integration/conftest.py
amandahla: Working on "Task was destroyed but it is pending!" even after successful integration tests.

src/charm.py
```diff
@@ -530,16 +528,15 @@ def _look_up_and_set_ingress_class(self, api, body):
...
-                    "They are: {}".format(" ".join(sorted(defaults)))
+                    f"They are: {default_ingress}"
                 )
                 return
 
             ingress_class = defaults[0]
-            LOGGER.info(
-                "Using ingress class {} as it is the cluster's default".format(ingress_class)
-            )
+            LOGGER.info(f"Using ingress class {ingress_class} as it is the cluster's default")
```
jdkandersson, line 539: Thank you for updating the code base to use f-strings, there is an exception to using f-strings in log messages as it is a security risk: https://github.com/canonical/is-charms-contributing-guide/blob/main/CONTRIBUTING.md#formatting-log-messages

tests/integration/conftest.py
```diff
@@ -50,47 +51,65 @@ async def app(ops_test: OpsTest, app_name: str):
...
-    # Get the IP address which is in the status message
+async def ip_address_list(ops_test: OpsTest, app: Application):
+    """Get unit IP address from workload message.
+
+    Example: Ingress IP(s): 127.0.0.1, Service IP(s): 10.152.183.84
+    """
+    # Reduce the update_status frequency until the cluster is deployed
+    async with ops_test.fast_forward():
+        await ops_test.model.block_until(
+            lambda: "Ingress IP(s)" in app.units[0].workload_status_message, timeout=100
```
jdkandersson, line 68: That is a nice implementation, I wasn't aware of that

jdkandersson:
> > I wonder why it looks like there is only 1 commit? Maybe it was the force push? Having multiple commits helps with making reviews easier on GitHub since reviewers can just see what has changed since their last review rather than reading the entire change again. We use squash merging anyway which means that in `main` it will end up looking like a single commit anyway
> 
> Something I do out of habit. The next ones will keep the commits.

Thank you for that, this follow up review was much easier as a result 😃

lib/charms/nginx_ingress_integrator/v0/ingress.py
```diff
@@ -18,6 +18,7 @@
 - limit-whitelist
 - max-body-size
 - owasp-modsecurity-crs
+- owasp-modsecurity-custom-rules
```
arturo-seijas, line 21: You'll need to increment the version of the library whenever you make changes so that charmcraft can identify those

arturo-seijas (changes requested):

tests/integration/conftest.py
amandahla: Possible fix here: https://github.com/canonical/charmed-magma-orchestrator/pull/44

amandahla: Didn't work. At least, it seems that doesn't have an impact on tests.

tests/integration/conftest.py
amandahla: I found while was researching about the "Task was destroyed..." error :)

src/charm.py
amandahla: Done :)

lib/charms/nginx_ingress_integrator/v0/ingress.py
amandahla: Not quite sure, like this?

```python
# Increment this major API version when introducing breaking changes
LIBAPI = 2

# Increment this PATCH version before using `charmcraft publish-lib` or reset
# to 0 if you are raising the major API version
LIBPATCH = 0
```

amandahla: LIBPATCH increased, thanks.

arturo-seijas (approved):
LGTM

weiiwang01:
> separate option for now that has no impact on owasp_modsecurity_crs and consider changing everything for just one option in the future.

Yes, we can come up with something in the future 👍

weiiwang01 (approved):

PR #40, opened 2022-11-24
Fix owasp-modsecurity-crs option in relation

weiiwang01:


tests/integration/test_charm.py
```diff
@@ -68,3 +75,85 @@ async def test_owasp_modsecurity_custom_rules(app_url_modsec_ignore: str):
...
+    arrange: given charm has been built and deployed
+    act: relate the ingress integrator with any-charm and use any-charm to toggle modsecurity
+        option through relation via ingress lib.
+    assert: modsecurity annotations should be attached and detached from the kubernetes ingress
+        resource according to the modsecurity option.
+    """
+    await ops_test.model.applications[app_name].set_config(
+        {"owasp-modsecurity-crs": "false", "owasp-modsecurity-custom-rules": ""}
+    )
+    any_charm_tmp_path = tmp_path / "any-charm"
```
arturo-seijas, line 91: Can I assume this charm is not it charmhub yet and this piece is temporary? The charms can be uploaded but not listed in charhub, which is probably what we want here

weiiwang01: I think we should do that in early test stages.

tests/integration/test_charm.py
```diff
@@ -68,3 +75,85 @@ async def test_owasp_modsecurity_custom_rules(app_url_modsec_ignore: str):
...
+    await ops_test.model.add_relation("any", f"{app_name}:ingress")
+    await ops_test.model.wait_for_idle(status=ActiveStatus.name)
+
+    kubernetes.config.load_kube_config()
+    kube = kubernetes.client.NetworkingV1Api()
+    ingress_annotations = kube.read_namespaced_ingress(
+        "any-ingress", namespace=ops_test.model.name
+    ).metadata.annotations
+    assert ingress_annotations["nginx.ingress.kubernetes.io/enable-modsecurity"] == "true"
+    assert (
```
merkata, line 137: Why is this assert in brackets whereas the other asserts aren't?

merkata (dismissed):
Looks great! Having any-charm at some point on charmhub would definetely be great along a short doc on how you can use it for testing, I really like it.

tests/integration/test_charm.py
weiiwang01: black use this trick to shorten the lines.

tests/integration/test_charm.py
arturo-seijas: Early as in "before merging" or as in "in the near future"? If the later, could you add a ticket to our backlog so we will recall updating this?

src/charm.py
```diff
@@ -155,7 +155,8 @@ def _max_body_size(self):
     @property
     def _owasp_modsecurity_crs(self):
         """Return a boolean indicating whether OWASP ModSecurity CRS is enabled."""
-        return self._get_config_or_relation_data("owasp-modsecurity-crs", False)
+        value = self._get_config_or_relation_data("owasp-modsecurity-crs", False)
+        return str(value).lower() == "true"
```
nrobinaubertin, line 159: Is `lower()` needed ? The latest yaml spec has only "true" and "false" as boolean scalars

weiiwang01: The ingress library will pass "True" and "False" in relation data, so it's needed here.

tests/integration/test_charm.py
weiiwang01: I will upload the `any-charm` before this being merged.

src/charm.py
amandahla: Add to BOOLEAN_CONFIG_FIELDS wouldn't' fix that?
Or this should be done for rewrite-enabled as well?

weiiwang01: It's already be done for `rewrite-enabled`, but you are right about adding this to `BOOLEAN_CONFIG_FIELDS`

tests/integration/test_charm.py
```diff
@@ -68,3 +75,85 @@ async def test_owasp_modsecurity_custom_rules(app_url_modsec_ignore: str):
...
+        class AnyCharm(AnyCharmBase):
+            def __init__(self, *args, **kwargs):
+                super().__init__(*args, **kwargs)
+                self.ingress = IngressRequires(
+                    self,
+                    {
+                        "service-hostname": "any",
+                        "service-name": self.app.name,
+                        "service-port": 80,
+                        "owasp-modsecurity-crs": True
```
gregory-schiano, line 114: One question here, which could be some miscomprehension on my side:
The charm uses the method `_get_config_or_relation_data` to get the value which will be use to insert in the ingress annontation
This `_get_config_or_relation_data` method will take the `nginx-ingress-integrator` charm config if it exists, or the value defined on the related charm (`any-charm` in our case here)
When I lookg at `config.yaml` of `ngninx-ingress-integrator` there's a default value to `false` for `owasp-modsecurity-crs` config...
So in my understanding what ever value is defined by the relation, since there's a config value to `false`, it will will be `false`

What do I miss ?

weiiwang01: The `owasp-modsecurity-crs` is not in `BOOLEAN_CONFIG_FIELDS`, so `_get_config` will return `None` if the `owasp-modsecurity-crs` set to `False`. So if `owasp-modsecurity-crs` config is `False` the modsecurity will depends on the relation. Is this a bug or a feature?

The more concerning part is `rewrite-enabled` which is out of this PR's scope. But `rewrite-enabled` is in the `BOOLEAN_CONFIG_FIELDS`, which makes it impossible to update that option via relation.

gregory-schiano: That's indeed what I missed. Well honestly I feel like this overriding rules "config overrides relation" or "relation overrides config" depending on the parameter are very confusing, and could lead to unexpected behavior from user perspective
I'm wondering if the charm shouldn't be much more opinionated and just forbid relation overriding the configuration, recent discussion with Tom and Arturo led to the conclusion that in most use cases there is 1 deploy ment of the nignx-ingress operator related to 1 backend charm (and not multiple backend related to it). Therefore there won't be consequences to only allow changes from the config side

weiiwang01: Yes, but it will break the backward compatibility, a major one. The charm rewrite would not be very difficult though.

tests/integration/test_charm.py
```diff
@@ -68,3 +75,85 @@ async def test_owasp_modsecurity_custom_rules(app_url_modsec_ignore: str):
...
+    if not any_charm_tmp_path.exists():
+        subprocess.run(  # nosec
+            ["git", "clone", "https://github.com/canonical/any-charm.git", any_charm_tmp_path]
+        )
+
+    any_charm = await ops_test.build_charm(any_charm_tmp_path)
+    ingress_lib = Path("lib/charms/nginx_ingress_integrator/v0/ingress.py").read_text()
+    any_charm_src_overwrite = {
+        "ingress.py": ingress_lib,
+        "any_charm.py": dedent(
```
jdkandersson, line 101: Another option here is to create a fixture file (like `any_charm.py`) and just read it from disk, it would mean that we get nice Python formatting when looking at the `any_charm.py` file and don't need to `dedent`. What do you think?

tests/integration/test_charm.py
```diff
@@ -68,3 +75,85 @@ async def test_owasp_modsecurity_custom_rules(app_url_modsec_ignore: str):
...
+        ),
+    }
+    await ops_test.model.deploy(
+        str(any_charm),
+        application_name="any",
+        series="focal",
+        config={"src-overwrite": json.dumps(any_charm_src_overwrite)},
+    )
+    await ops_test.model.add_relation("any", f"{app_name}:ingress")
+    await ops_test.model.wait_for_idle(status=ActiveStatus.name)
```
jdkandersson, line 129: Potentially the code up to here could be a fixture? I guess it depends on if it is used in other tests/ if it helps with readability

tests/integration/test_charm.py
```diff
@@ -68,3 +75,85 @@ async def test_owasp_modsecurity_custom_rules(app_url_modsec_ignore: str):
...
+    )
+    assert ingress_annotations["nginx.ingress.kubernetes.io/modsecurity-snippet"]
+
+    action = (
+        await ops_test.model.applications["any"]
+        .units[0]
+        .run_action(
+            "rpc",
+            method="update_ingress",
+            kwargs=json.dumps({"ingress_config": {"owasp-modsecurity-crs": False}}),
```
jdkandersson, line 148: Excellent, the `any-charm` is quite helpful here!

tests/integration/test_charm.py
weiiwang01: Yes, I will put it in a separated file.

tests/integration/test_charm.py
weiiwang01: It's just used once here, maybe we can put this in a fixture when we add more integration tests in the future.

arturo-seijas (approved):
LGTM

amandahla (approved):

src/charm.py
weiiwang01: Hi Amanda, see the discussion down below, I think I will leave this problem to the future.

PR #41, opened 2022-11-28
Revert to publishable library version

mthaddon:
The previously published version of the library was 0.10. This set it to 0.11 which can be published (have already done so).

This means no-one needs to update charm code to get the new version, they can simply update the existing v0.

### DIFF
--- a/lib/charms/nginx_ingress_integrator/v0/ingress.py
+++ b/lib/charms/nginx_ingress_integrator/v0/ingress.py
@@ -65,11 +65,11 @@ from ops.model import BlockedStatus
 LIBID = "db0af4367506491c91663468fb5caa4c"
 
 # Increment this major API version when introducing breaking changes
-LIBAPI = 1
+LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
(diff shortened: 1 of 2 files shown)


gtrkiller (commented):
LGTM. Do we want to pin flake8 & pyproject-flake8 to <6.0.0 temporarily for the lint CI/CD to pass? or leave it as it is?

jdkandersson (approved):

jdkandersson:
Did they fix the limitation that the minor version could not be more than 9?

mthaddon:
> Did they fix the limitation that the minor version could not be more than 9?

It turns out there isn't a limitation on the minor version, but we were trying to publish a non-sequential version number. The error message is being updated to clarify this.

mthaddon:
> LGTM. Do we want to pin flake8 & pyproject-flake8 to <6.0.0 temporarily for the lint CI/CD to pass? or leave it as it is?

Yeah, I think it's worth doing for now, have added that change.

gregory-schiano (approved):
LGTM

arturo-seijas (approved):
LGTM

PR #45, opened 2023-01-13
try out migration

jdkandersson:
Test PR, please do not merge, no need to review

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -5,5 +5,5 @@ on:
 
 jobs:
   unit-tests:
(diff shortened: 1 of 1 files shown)


jdkandersson:
@mthaddon would you mind updating the discourse secrets on this repo with your username and API access token? It looks like the posts are owned by you which means that migration can only work if the API key has write permission. Or maybe we make these wiki posts that anyone can edit? I wonder if that would work in this case. The topic that caused the error: https://discourse.charmhub.io/t/nginx-ingress-integrator-docs-tutorial-adding-relation-to-a-charm/7434

PR #49, opened 2023-01-23
fix(ISD-359): Add kubernetes library version to status message

amandahla:
nginx-ingress-integrator reports kubernetes livrary version in juju status

Example (nginx1):

image

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -768,7 +768,16 @@ class NginxIngressCharm(CharmBase):
                     "Duplicate route found; cannot add ingress. Run juju debug-log for details."
                 )
                 return
+        self.unit.set_workload_version(self._get_kubernetes_library_version())
         self.unit.status = ActiveStatus(msg)
+
+    def _get_kubernetes_library_version(self) -> str:
+        """Retrieve the current version of Kubernetes library.
+
+        Returns:
+            The Kubernetes library used.
+        """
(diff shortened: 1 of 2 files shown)


jdkandersson:
Would it be worth adding why this is needed? Was this requested by users?

src/charm.py
```diff
@@ -768,8 +768,17 @@ def _on_config_changed(self, _):
                     "Duplicate route found; cannot add ingress. Run juju debug-log for details."
                 )
                 return
+        self.unit.set_workload_version(self._get_kubernetes_library_version())
```
jdkandersson, line 771: I wonder what the trade-offs would be for putting `kubernetes.__version__` here instead of creating another function?

jdkandersson, line 771: Should any tests be modified to ensure the version shows up as expected in the status message?

amandahla: The story had a reference to Indico and this is how was done there as you can see [here](https://github.com/canonical/indico-operator/blob/main/src/charm.py#L262). I thought would be nice to follow the same pattern. Also, if by any chance would be needed to format the version string, it would be easier.

amandahla: Good catch. Unit tests updated.

arturo-seijas (approved):
LGTM

merkata (approved):
Nice!

PR #50, opened 2023-01-26
fix(ISD-362): Remove unused resources

### LINKED ISSUE #4: Leaked Kubernetes resources when a service-name / service-hostname changes
The nginx-ingress-integrator charm creates Kubernetes resources (Services and Ingress Resources) based on the configured service-name and service-hostname.

However, if the charm is then configured with a different service-name and service-hostname, the previous resources are not removed. This happens everytime the configuration options change since the Resource names are based on the aforementioned service-name / service-hostname.

This can also happen for the service-names and service-hostnames received from relation data (+ additional-hostnames for Ingress Resources). This issue occurs on previous revisions as well.

Because the Kubernetes resource names are based on service-name / service-hostname, cleaning them up is not immediately straight-forward since the charm would have to store the previous values and compare them to the current ones.

amandahla:
Fix #4 

This PR:
- Add a mechanism that for every event, checks if there are unused services or ingress and delete them. 
This is done by:
1) getting all services/ingress resources with the label "app.juju.is/created-by=".
2) getting all services/ingress resources from config or relations 
3) Each service/ingress resource that is not in the list mentioned above is deleted.
- Add unit and integration tests
- Add mypy checker and related fixes or disabled types
- Change AnyCharm to get Ingress values from the relation, if exists. The default values were set in the _\_init_\_ and since charm is a stateless application, every event was initializing the charm and recreating the ingress with the default values. Reference [here](https://juju.is/docs/sdk/create-a-minimal-kubernetes-charm#heading--define-the-charm-initialisation-and-application-services).

Findings that will be addressed in further PRs:
- Custom label for NGINX charm (to be confirmed)
- Refactoring of unit tests (ISD-417)
- Refactoring of Type hints (ISD-422)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,6 +14,9 @@ branch = true
 fail_under = 93
 show_missing = true
 
+[tool.mypy]
+ignore_missing_imports = true
+allow_redefinition = true
 
 [tool.pytest.ini_options]
 minversion = "6.0"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,12 +15,12 @@ fail_under = 93
 show_missing = true
 
 [tool.mypy]
-ignore_missing_imports = true
-allow_redefinition = true
+check_untyped_defs = true
+disallow_untyped_defs = true
 
-[tool.pytest.ini_options]
-minversion = "6.0"
-log_cli_level = "INFO"
+[[tool.mypy.overrides]]
+module = "tests.*"
+disallow_untyped_defs = false
 
 # Formatting tools configuration
 [tool.black]
--- a/src/charm.py
+++ b/src/charm.py
@@ -802,7 +802,23 @@ class NginxIngressCharm(CharmBase):
                 ingress_name,
             )
 
-    def _on_config_changed(self, _):
+    def _delete_unused_resources(self) -> None:
+        """Delete unused services and ingresses"""
+        svc_names = [conf_or_rel._service_name for conf_or_rel in self._all_config_or_relations]
+        self._delete_unused_services(svc_names)
+        svc_hostnames = [
+            conf_or_rel._service_hostname for conf_or_rel in self._all_config_or_relations
+        ]
+        all_additional_hostnames = [
+            conf_or_rel._additional_hostnames
+            for conf_or_rel in self._all_config_or_relations
+        ]
+        for additional_hostname in all_additional_hostnames:
+            svc_hostnames.extend(additional_hostname)
+        self._delete_unused_ingresses(svc_hostnames)
+
+
+    def _on_config_changed(self, event):
         """Handle the config changed event."""
         msg = ""
         # We only want to do anything here if we're the leader to avoid
@@ -818,17 +834,7 @@ class NginxIngressCharm(CharmBase):
                     msgs.append(f"Ingress IP(s): {', '.join(ingress_ips)}")
                 msgs.append(f"Service IP(s): {', '.join(self._report_service_ips())}")
                 msg = ", ".join(msgs)
-                self._delete_unused_services(svc_names)
-                svc_hostnames = [
-                    conf_or_rel._service_hostname for conf_or_rel in self._all_config_or_relations
-                ]
-                all_additional_hostnames = [
-                    conf_or_rel._additional_hostnames
-                    for conf_or_rel in self._all_config_or_relations
-                ]
-                for additional_hostname in all_additional_hostnames:
-                    svc_hostnames.extend(additional_hostname)
-                self._delete_unused_ingresses(svc_hostnames)
+                self._delete_unused_resources()
             except kubernetes.client.exceptions.ApiException as exception:
                 if exception.status == 403:
                     LOGGER.error(
--- a/src/charm.py
+++ b/src/charm.py
@@ -23,6 +23,9 @@ from ops.model import ActiveStatus, BlockedStatus
 LOGGER = logging.getLogger(__name__)
 _INGRESS_SUB_REGEX = re.compile("[^0-9a-zA-Z]")
 BOOLEAN_CONFIG_FIELDS = ["rewrite-enabled"]
+# Juju defines the value of this label.
+# It has the same value as the label "app.kubernetes.io/name"
+# set in the service account associated with the application.
 CREATED_BY_LABEL = "app.juju.is/created-by="
 
 
--- a/src/charm.py
+++ b/src/charm.py
@@ -28,6 +28,8 @@ BOOLEAN_CONFIG_FIELDS = ["rewrite-enabled"]
 # Juju defines the value of this label.
 # It has the same value as the label "app.kubernetes.io/name"
 # set in the service account associated with the application.
+KUBERNETES_CORE_API = "core"
+KUBERNETES_NETWORKING_API = "networking"
 CREATED_BY_LABEL = "app.juju.is/created-by="
 
 
@@ -428,9 +430,7 @@ class NginxIngressCharm(CharmBase):
 
     def _describe_ingresses_action(self, event):
         """Handle the 'describe-ingresses' action."""
-        self.k8s_auth()
-        api = _networking_v1_api()
-
+        api = self.k8s_api(api=KUBERNETES_NETWORKING_API)
         ingresses = api.list_namespaced_ingress(namespace=self._namespace)
         event.set_results({"ingresses": ingresses})
 
@@ -443,11 +443,26 @@ class NginxIngressCharm(CharmBase):
 
         self._authed = True
 
+    def k8s_api(self, api: str) -> object:
+        """Authenticate and return a k8s API.
+
+        Args:
+            api: Which API should be returned. If is not found, None is returned.
+        """
+        self.k8s_auth()
+        if api == KUBERNETES_CORE_API:
+            return _core_v1_api()
+        if api == KUBERNETES_NETWORKING_API:
+            return _networking_v1_api()
+        LOGGER.error("k8s api %s not found", api)
+        return None
+
     def _report_service_ips(self) -> List[str]:
         """Report on service IP(s) and return a list of them."""
-        self.k8s_auth()
-        api = _core_v1_api()
-        services = api.list_namespaced_service(namespace=self._namespace)
+        api = self.k8s_api(KUBERNETES_CORE_API)
+        services = api.list_namespaced_service(  # type: ignore[attr-defined]
+            namespace=self._namespace
+        )
         all_k8s_service_names = [rel._k8s_service_name for rel in self._all_config_or_relations]
         return [
             x.spec.cluster_ip for x in services.items if x.metadata.name in all_k8s_service_names
@@ -455,13 +470,13 @@ class NginxIngressCharm(CharmBase):
 
     def _report_ingress_ips(self) -> List[str]:
         """Report on ingress IP(s) and return a list of them."""
-        self.k8s_auth()
-        api = _networking_v1_api()
-
+        api = self.k8s_api(KUBERNETES_NETWORKING_API)
         # Wait up to `interval * count` seconds for ingress IPs.
         count, interval = 100, 1
         for _ in range(count):
-            ingresses = api.list_namespaced_ingress(namespace=self._namespace)
+            ingresses = api.list_namespaced_ingress(  # type: ignore[attr-defined]
+                namespace=self._namespace
+            )
             try:
                 ips = [x.status.load_balancer.ingress[0].ip for x in ingresses.items]
             except TypeError:
@@ -479,6 +494,9 @@ class NginxIngressCharm(CharmBase):
         field_names = [f'_{f.replace("-", "_")}' for f in REQUIRED_INGRESS_RELATION_FIELDS]
         return all(getattr(conf_or_rel, f) for f in field_names)
 
+    def _get_created_label(self) -> str:
+        return f"{CREATED_BY_LABEL}{self.app.name}"
+
     def _delete_unused_services(self, current_svc_names: List[str]) -> None:
         """Delete services and ingresses that are no longer used.
 
@@ -486,11 +504,9 @@ class NginxIngressCharm(CharmBase):
             current_svc_names: service names set by config or relation data
             current_svc_hostnames: service hostnames set by config or relation data
         """
-        self.k8s_auth()
-        api = _core_v1_api()
-        created_by_label = f"{CREATED_BY_LABEL}{self.app.name}"
-        all_services = api.list_namespaced_service(
-            namespace=self._namespace, label_selector=created_by_label
+        api = self.k8s_api(KUBERNETES_CORE_API)
+        all_services = api.list_namespaced_service(  # type: ignore[attr-defined]
+            namespace=self._namespace, label_selector=self._get_created_label()
         )
         all_svc_names = tuple(item.metadata.name for item in all_services.items)
         unused_svc_names = tuple(
@@ -505,7 +521,7 @@ class NginxIngressCharm(CharmBase):
             unused_svc_names,
         )
         for unused_svc_name in unused_svc_names:
-            api.delete_namespaced_service(
+            api.delete_namespaced_service(  # type: ignore[attr-defined]
                 name=unused_svc_name,
                 namespace=self._namespace,
             )
@@ -526,12 +542,13 @@ class NginxIngressCharm(CharmBase):
 
     def _define_service(self, conf_or_rel: _ConfigOrRelation):
         """Create or update a service in kubernetes."""
-        self.k8s_auth()
-        api = _core_v1_api()
+        api = self.k8s_api(KUBERNETES_CORE_API)
         body = conf_or_rel._get_k8s_service()
-        services = api.list_namespaced_service(namespace=self._namespace)
+        services = api.list_namespaced_service(  # type: ignore[attr-defined]
+            namespace=self._namespace
+        )
         if conf_or_rel._k8s_service_name in [x.metadata.name for x in services.items]:
-            api.patch_namespaced_service(
+            api.patch_namespaced_service(  # type: ignore[attr-defined]
                 name=conf_or_rel._k8s_service_name,
                 namespace=self._namespace,
                 body=body,
@@ -542,7 +559,7 @@ class NginxIngressCharm(CharmBase):
                 conf_or_rel._service_name,
             )
         else:
-            api.create_namespaced_service(
+            api.create_namespaced_service(  # type: ignore[attr-defined]
                 namespace=self._namespace,
                 body=body,
             )
@@ -554,11 +571,12 @@ class NginxIngressCharm(CharmBase):
 
     def _remove_service(self, conf_or_rel: _ConfigOrRelation):
         """Remove the created service in kubernetes."""
-        self.k8s_auth()
-        api = _core_v1_api()
-        services = api.list_namespaced_service(namespace=self._namespace)
+        api = self.k8s_api(KUBERNETES_CORE_API)
+        services = api.list_namespaced_service(  # type: ignore[attr-defined]
+            namespace=self._namespace
+        )
         if conf_or_rel._k8s_service_name in [x.metadata.name for x in services.items]:
-            api.delete_namespaced_service(
+            api.delete_namespaced_service(  # type: ignore[attr-defined]
                 name=conf_or_rel._k8s_service_name,
                 namespace=self._namespace,
             )
@@ -604,11 +622,9 @@ class NginxIngressCharm(CharmBase):
         Args:
             current_svc_hostnames: service hostnames set by config or relation data
         """
-        self.k8s_auth()
-        api = _networking_v1_api()
-        created_by_label = f"{CREATED_BY_LABEL}{self.app.name}"
-        all_ingresses = api.list_namespaced_ingress(
-            namespace=self._namespace, label_selector=created_by_label
+        api = self.k8s_api(KUBERNETES_NETWORKING_API)
+        all_ingresses = api.list_namespaced_ingress(  # type: ignore[attr-defined]
+            namespace=self._namespace, label_selector=self._get_created_label()
         )
         all_svc_hostnames = tuple(ingress.spec.rules[0].host for ingress in all_ingresses.items)
         unused_svc_hostnames = tuple(
@@ -766,8 +782,7 @@ class NginxIngressCharm(CharmBase):
 
     def _define_ingress(self, body):
         """Create or update an ingress in kubernetes."""
-        self.k8s_auth()
-        api = _networking_v1_api()
+        api = self.k8s_api(KUBERNETES_NETWORKING_API)
         self._look_up_and_set_ingress_class(api, body)
         ingress_name = body.metadata.name
         ingresses = api.list_namespaced_ingress(namespace=self._namespace)
@@ -795,8 +810,7 @@ class NginxIngressCharm(CharmBase):
 
     def _remove_ingress(self, ingress_name):
         """Remove ingress resource."""
-        self.k8s_auth()
-        api = _networking_v1_api()
+        api = self.k8s_api(KUBERNETES_NETWORKING_API)
         ingresses = api.list_namespaced_ingress(namespace=self._namespace)
         if ingress_name in [x.metadata.name for x in ingresses.items]:
             api.delete_namespaced_ingress(ingress_name, self._namespace)
--- a/src/charm.py
+++ b/src/charm.py
@@ -9,7 +9,7 @@
 import logging
 import re
 import time
-from typing import List
+from typing import Generator, List
 
 import kubernetes.client
 from charms.nginx_ingress_integrator.v0.ingress import (
@@ -121,14 +121,16 @@ class _ConfigOrRelation:
         return fallback
 
     @property
-    def _additional_hostnames(self) -> List[str]:
+    def _additional_hostnames(self) -> Generator[str, None, None]:
         """Return a list with additional hostnames.
 
         Returns:
             The additional hostnames set by configuration already split by comma.
         """
         additional_hostnames = self._get_config_or_relation_data("additional-hostnames", "")
-        return [hostname for hostname in additional_hostnames.split(",") if hostname]
+        for hostname in additional_hostnames.split(","):
+            if hostname:
+                yield hostname
 
     @property
     def _k8s_service_name(self):
--- a/src/charm.py
+++ b/src/charm.py
@@ -116,6 +116,19 @@ class _ConfigOrRelation:
         return fallback
 
     @property
+    def _additional_hostnames(self) -> List[str]:
+        """Return a list with additional hostnames.
+
+        Returns:
+            The additional hostnames set by configuration already split by comma.
+        """
+        return [
+            x
+            for x in self._get_config_or_relation_data("additional-hostnames", "").split(",")
+            if x
+        ]
+
+    @property
     def _k8s_service_name(self):
         """Return a service name for the use creating a k8s service."""
         # Avoid collision with service name created by Juju. Currently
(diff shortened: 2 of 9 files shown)


src/charm.py
```diff
@@ -467,6 +475,42 @@ def _has_required_fields(self, conf_or_rel: _ConfigOrRelation):
...
+        all_services = api.list_namespaced_service(
+            namespace=self._namespace, label_selector=created_by_label
+        )
+        all_svc_names = [item.metadata.name for item in all_services.items]
+        unused_svc_names = [
+            svc_name
+            for svc_name in all_svc_names
+            if svc_name.replace("-service", "") not in current_svc_names
+        ]
+        LOGGER.info(
```
gregory-schiano, line 497: I think this one should be debug

src/charm.py
```diff
@@ -550,6 +594,31 @@ def _look_up_and_set_ingress_class(self, api, body):
...
+        api = _networking_v1_api()
+        created_by_label = f"{CREATED_BY_LABEL}{self.app.name}"
+        all_ingresses = api.list_namespaced_ingress(
+            namespace=self._namespace, label_selector=created_by_label
+        )
+        all_svc_hostnames = [ingress.spec.rules[0].host for ingress in all_ingresses.items]
+        unused_svc_hostnames = [
+            hostname for hostname in all_svc_hostnames if hostname not in current_svc_hostnames
+        ]
+        LOGGER.info(
```
gregory-schiano, line 613: Same as above

src/charm.py
```diff
@@ -23,6 +23,7 @@
 LOGGER = logging.getLogger(__name__)
 _INGRESS_SUB_REGEX = re.compile("[^0-9a-zA-Z]")
 BOOLEAN_CONFIG_FIELDS = ["rewrite-enabled"]
+CREATED_BY_LABEL = "app.juju.is/created-by="
```
gregory-schiano, line 26: As discussed in the daily digest, I don't understand where this label comes from.
The charm uses python kubernetes-client library and doesn't explicitly set this label which means:
- Either the library takes it from the pod calling the API (I don't think that's the case, I quickly took a look at the library code and didn't find such business logic)
- Or it's added by Juju for every ressource within the namespace (perhaps edited automatically by the modelcontroller ? But how does juju show who created this kubernetes ressources as there could be multiple charms deployed in a model)

This should be clarified for 2 reasons:
- Ensure we won't delete services/ingress we don't want to delete
- Ensure this label won't be removed in a future releases

One approach could be to add our own label in this charm when creating the Ingress/Service, the drawback is that we won't be able to cleanup previously created ressources (in case of an upgrade of charm)

amandahla: After some digging, I think I found out how Juju knows which application created the resource.

When Juju receives a request via cassadmission, fetches the juju application name associated with a service account that in its turn has this value set in the label "app.kubernetes.io/name".
 
Code:
https://github.com/juju/juju/blob/develop/worker/caasadmission/handler.go#L131

microk8s output:
```
$ microk8s kubectl get serviceaccount -n indico nginx1 --show-labels
NAME     SECRETS   AGE     LABELS
nginx1   0         4h48m   app.kubernetes.io/managed-by=juju,app.kubernetes.io/name=nginx1
$ microk8s kubectl get ing -n indico --show-labels
NAME                    CLASS    HOSTS           ADDRESS     PORTS   AGE     LABELS
indico-local-ingress    public   indico.local    127.0.0.1   80      4h46m   app.juju.is/created-by=nginx1
hello-kubecon-ingress   public   hello-kubecon   127.0.0.1   80      6m5s    app.juju.is/created-by=nginx2
```

src/charm.py
```diff
@@ -299,13 +313,7 @@ def _get_k8s_ingress(self):
...
 
         hostnames = [self._service_hostname]
-        hostnames.extend(
-            [
-                x
-                for x in self._get_config_or_relation_data("additional-hostnames", "").split(",")
-                if x
-            ]
-        )
+        hostnames.extend(self._additional_hostnames)
```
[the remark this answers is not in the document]
amandahla: No, it's used in _on_config_changed too.

src/charm.py
```diff
@@ -114,6 +115,19 @@ def _get_relation_data_or_config(self, field, fallback):
...
         return fallback
 
+    @property
+    def _additional_hostnames(self) -> List[str]:
+        """Return a list with additional hostnames.
+
+        Returns:
+            The additional hostnames set by configuration already split by comma.
+        """
+        return [
```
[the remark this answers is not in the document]
amandahla: Thanks for pointing that out but since one of the use cases is to check if a hostname is in the list (current hostname + all additional hostnames), an iterator wouldn't be a good fit.

src/charm.py
```diff
@@ -467,6 +475,42 @@ def _has_required_fields(self, conf_or_rel: _ConfigOrRelation):
...
+            current_svc_names: service names set by config or relation data
+            current_svc_hostnames: service hostnames set by config or relation data
+        """
+        self.k8s_auth()
+        api = _core_v1_api()
+        created_by_label = f"{CREATED_BY_LABEL}{self.app.name}"
+        all_services = api.list_namespaced_service(
+            namespace=self._namespace, label_selector=created_by_label
+        )
+        all_svc_names = [item.metadata.name for item in all_services.items]
```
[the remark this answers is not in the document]
amandahla: Do you think would be worth using a generator instead of a list comprehension for a small list? I'm assuming that would be rare to be more than 100 items or even 50 in this case.

src/charm.py
```diff
@@ -550,6 +594,31 @@ def _look_up_and_set_ingress_class(self, api, body):
...
 
+    def _delete_unused_ingresses(self, current_svc_hostnames: List[str]):
+        """Delete ingresses that are no longer managed by nginx-ingress-integrator charm.
+
+        Args:
+            current_svc_hostnames: service hostnames set by config or relation data
+        """
+        self.k8s_auth()
+        api = _networking_v1_api()
+        created_by_label = f"{CREATED_BY_LABEL}{self.app.name}"
```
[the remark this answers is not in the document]
amandahla: They use different APIs :(

src/charm.py
```diff
@@ -745,6 +813,17 @@ def _on_config_changed(self, _):
                     msgs.append(f"Ingress IP(s): {', '.join(ingress_ips)}")
                 msgs.append(f"Service IP(s): {', '.join(self._report_service_ips())}")
                 msg = ", ".join(msgs)
+                self._delete_unused_services(svc_names)
+                svc_hostnames = [
```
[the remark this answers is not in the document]
amandahla: They are separated because in this way one function does just one thing and would be easier to modify/debug if needed. Another reason is that they use different API calls.

src/charm.py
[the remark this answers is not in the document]
amandahla: (I'm asking out of curiosity, I applied your suggestion)

tests/integration/test_relation.py
```diff
@@ -51,6 +53,38 @@ async def build_and_deploy_ingress():
...
+    model_name = ops_test.model.name
+    created_by_label = f"{CREATED_BY_LABEL}ingress"
+    all_ingresses = api_networking.list_namespaced_ingress(
+        namespace=model_name, label_selector=created_by_label
+    )
+    all_svc_hostnames = [ingress.spec.rules[0].host for ingress in all_ingresses.items]
+    assert all_svc_hostnames == ["any"]
+    await ops_test.juju("config", INGRESS_APP_NAME, "service-hostname=new-name")
+    await ops_test.model.wait_for_idle(status="active")
+    all_ingresses = api_networking.list_namespaced_ingress(
```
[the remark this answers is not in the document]
amandahla: Good catch.

tests/unit/test_charm.py
```diff
@@ -20,6 +20,8 @@ def setUp(self):
         self.addCleanup(self.harness.cleanup)
         self.harness.begin()
 
+    @patch("charm.NginxIngressCharm._delete_unused_ingresses")
```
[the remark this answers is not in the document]
amandahla: Thanks for sharing the blog post. Since would be out of the scope of this PR, I added the autospec parameter only to the mocks created by this PR. I believe we could open a new story for refactoring the tests, what do you think?

src/charm.py
gregory-schiano: Ah makes sens, it uses an Admission Controller to get the events, and uses the Service Account attached to the pod.
Thx for digging, good to know anyway :)

Did you get some feedback from Juju team on this topic ?

src/charm.py
```diff
@@ -123,9 +123,11 @@ def _additional_hostnames(self) -> List[str]:
             The additional hostnames set by configuration already split by comma.
         """
         return [
-            x
-            for x in self._get_config_or_relation_data("additional-hostnames", "").split(",")
-            if x
+            hostname
+            for hostname in self._get_config_or_relation_data("additional-hostnames", "").split(
+                ","
+            )
```
gregory-schiano, line 132: The indentation seems strange, wouldn't it be better to return to the line before ".split" ?

src/charm.py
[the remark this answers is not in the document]
jdkandersson: That is on the user side, it doesn't stop you from returning an iterator and the caller passing it to the list constructor. By returning a list, the use of a list is forced rather than up to the caller. The caller could be just interested in the first one or just need to process it

src/charm.py
```diff
@@ -467,6 +475,42 @@ def _has_required_fields(self, conf_or_rel: _ConfigOrRelation):
         field_names = [f'_{f.replace("-", "_")}' for f in REQUIRED_INGRESS_RELATION_FIELDS]
         return all(getattr(conf_or_rel, f) for f in field_names)
 
+    def _delete_unused_services(self, current_svc_names: List[str]):
```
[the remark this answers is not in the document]
jdkandersson: Was the mypy configuration added? I can't spot the change in this PR?

src/charm.py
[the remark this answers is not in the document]
jdkandersson: If a generator works, why not use it? It is more efficient (has less functionality than a list) and uses less memory

src/charm.py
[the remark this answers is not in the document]
jdkandersson: Take the API is a parameter?

src/charm.py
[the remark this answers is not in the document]
jdkandersson: What about adding another function that removes all the unused resources rather than writing that in the `_on_config_changed` code?

tests/integration/test_relation.py
```diff
@@ -51,6 +53,38 @@ async def build_and_deploy_ingress():
...
 
 
+@pytest.mark.usefixtures("build_and_deploy")
+async def test_delete_unused_ingresses(ops_test: OpsTest):
+    """
+    arrange: given charm has been built, deployed and related to a dependent application
+    act: when the service-hostname is changed and when is back to previous value
+    assert: then the workload status is active and the unused ingress is deleted
+    """
+    kubernetes.config.load_kube_config()
```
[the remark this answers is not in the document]
jdkandersson: I meant in the code itself, i.e., after the block of code that does the arrange, add an empty line. Looks a bit better now anyway

tests/unit/test_charm.py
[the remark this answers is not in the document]
jdkandersson: Yeh it is probably reasonable to do so, although this is kind of digging a deeper hole. The use of mocks her looks excessive which could indicate that the unit tests should actually be integration tests or that we need to write a few abstractions for the external dependencies that are easier to mock. Any thoughts @gregory-schiano?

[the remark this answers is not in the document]
gregory-schiano: Yeap perhaps doing like Weii did on Spring Boot Charm, creating some mock classes 
https://github.com/canonical/spring-boot-k8s-operator/blob/main/tests/unit/spring_boot_patch.py

Globally speaking if we start needing more mock classes like this, it could be helpful to make a "Charm Mock module" in a separate repo we could easily import and use instead of defining those in multiple charms

src/charm.py
amandahla: Yes, here it is:
https://chat.charmhub.io/charmhub/pl/3ik6j6ryrjby7ykjx8fmy9keha

```
I can't guarantee these won't change in the future. We'll try to ensure they don't get changed, but I would recommend implementing a test on your end.

Why I could see this changing at some point is two things, labelling of resources created by the app isn't perfect, I think there is still some edge cases to work through; secondly the app name isn't unique, so we need to add a model-uuid into it, since resources could be created outside the model managed namespace.
```

src/charm.py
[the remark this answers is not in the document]
amandahla: Sorry, I just added the "-> None". Next commit will have mypy.

src/charm.py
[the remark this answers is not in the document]
amandahla: Done.

src/charm.py
amandahla: Sure, changed. :)

src/charm.py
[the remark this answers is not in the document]
amandahla: BTW I added a lot of "type: ignore"/"disable"s but without comment because would be repetitive. I'm not sure if this is the best approach.
There are many cases like:
```
tests/integration/conftest.py:40: error: Item "None" of "Optional[Model]" has no attribute "deploy"  [union-attr]
```

src/charm.py
amandahla: We could work with both options. First, we could look for our label. If doesn't exist then we look for the app.juju.is/created-by=. If none of them exist, then nothing will be deleted. What do you think?

src/charm.py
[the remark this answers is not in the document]
amandahla: Changed :)

src/charm.py
```diff
@@ -745,6 +813,17 @@ def _on_config_changed(self, _):
...
                 msg = ", ".join(msgs)
+                self._delete_unused_services(svc_names)
+                svc_hostnames = [
+                    conf_or_rel._service_hostname for conf_or_rel in self._all_config_or_relations
+                ]
+                all_additional_hostnames = [
+                    conf_or_rel._additional_hostnames
+                    for conf_or_rel in self._all_config_or_relations
+                ]
+                for additional_hostname in all_additional_hostnames:
```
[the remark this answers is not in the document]
amandahla: Moved to self._delete_unused_resources()

src/charm.py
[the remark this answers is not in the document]
jdkandersson: So for those just put an `assert app.model` or whatever at the start of the test and they will go away. Mypy is telling you that `ops` doesn't guarantee the model is defined - therefore we should check that it is if we are going to use it

pyproject.toml
```diff
@@ -14,6 +14,9 @@ branch = true
 fail_under = 93
 show_missing = true
 
+[tool.mypy]
```
jdkandersson, line 17: Specifically, the contributing guide asks for this:

```tox
[tool.mypy]
check_untyped_defs = true
disallow_untyped_defs = true

[[tool.mypy.overrides]]
module = "tests.*"
disallow_untyped_defs = false
```

src/charm.py
```diff
@@ -116,17 +123,16 @@ def _get_relation_data_or_config(self, field, fallback):
...
         Returns:
             The additional hostnames set by configuration already split by comma.
         """
-        return [
-            x
-            for x in self._get_config_or_relation_data("additional-hostnames", "").split(",")
-            if x
-        ]
+        additional_hostnames = self._get_config_or_relation_data("additional-hostnames", "")
+        for hostname in additional_hostnames.split(","):
```
jdkandersson, line 133: You could do this by writing: `yield from filter(None, additional_hostnames.split(","))`

src/charm.py
```diff
@@ -439,25 +443,40 @@ def k8s_auth(self):
...
+            return _networking_v1_api()
+        LOGGER.error("k8s api %s not found", api)
+        return None
+
     def _report_service_ips(self) -> List[str]:
         """Report on service IP(s) and return a list of them."""
-        self.k8s_auth()
-        api = _core_v1_api()
-        services = api.list_namespaced_service(namespace=self._namespace)
+        api = self.k8s_api(KUBERNETES_CORE_API)
```
jdkandersson, line 462: Is this an improvement? Maybe we have a function `core_v1_api` that returns the `core_v1_api` after calling `k8s_auth`? We probably lose a lot of type hint value by making the api to use generic

src/charm.py
```diff
@@ -699,7 +719,12 @@ def _process_ingresses(self, ingresses):
                 # Ensure that the exact same route for this route was not yet defined.
                 defined_paths = ingress_paths[rule.host]
                 for path in rule.http.paths:
-                    duplicate_paths = list(filter(lambda p: p.path == path.path, defined_paths))
+                    duplicate_paths = list(
+                        filter(
+                            lambda p: p.path == path.path,  # pylint: disable=cell-var-from-loop
```
jdkandersson, line 724: SHould we note why this disable is reasonable in this case please?

src/charm.py
```diff
@@ -813,17 +850,7 @@ def _on_config_changed(self, _):
...
-                    conf_or_rel._service_hostname for conf_or_rel in self._all_config_or_relations
-                ]
-                all_additional_hostnames = [
-                    conf_or_rel._additional_hostnames
-                    for conf_or_rel in self._all_config_or_relations
-                ]
-                for additional_hostname in all_additional_hostnames:
-                    svc_hostnames.extend(additional_hostname)
-                self._delete_unused_ingresses(svc_hostnames)
+                self._delete_unused_resources()
```
jdkandersson, line 853: Nice!

src/charm.py
```diff
@@ -699,7 +719,12 @@ def _process_ingresses(self, ingresses):
                 # Ensure that the exact same route for this route was not yet defined.
                 defined_paths = ingress_paths[rule.host]
                 for path in rule.http.paths:
-                    duplicate_paths = list(filter(lambda p: p.path == path.path, defined_paths))
+                    duplicate_paths = list(
```
jdkandersson, line 722: This function looks like it would benefit from refactoring, there are a lot of for loops in for loops, if statement etc

src/charm.py
jdkandersson: This can also be done without a lambda: `duplicate_paths = [defined_path for defined_path in defined_paths if defined_path.path == path.path]`

Writing this I'm confused by all the paths, the variable names here could be improved for both `defined_paths` and `path` since both of them have a `path` attribute which means that the outer variable probably is more than just a `path`

src/charm.py
jdkandersson: Or yield from `(hostname for hostname in additional_hostnames if hostname)` depending on preference

src/charm.py
```diff
@@ -431,25 +443,40 @@ def k8s_auth(self):
 
         self._authed = True
 
+    def k8s_api(self, api: str) -> object:
```
jdkandersson, line 446: I'm curios what the reason was for adding this?

src/charm.py
merkata: +1 for rephrasing, all this "path path, path" is rhythmic, though it's better suited in a song and not in code.

src/charm.py
gregory-schiano: Well not sure the added business logic worth it
- Having our own label, will never trigger the fallback to "created-by", unless someone removes it from our code. In that case we should ensure a unit/integration test will fail due to broken feature
- Relying on juju "created-by" label, will only break if they remove it, and here once more we should have a unit/integration tests that will fail if that's the case.

I'm more in favor of the 1st one because we don't run our integration tests scheduled, which means that the feature could be broken without us noticing until we do a code change that breaks the integration tests.

src/charm.py
amandahla: I might misunderstand one of your reviews when you said "Take the API is a parameter?" so I created this method receiving the wanted API as a parameter. I'll review if it's ok to change every call to core_v1_api/networking to include self.k8s_auth instead.

src/charm.py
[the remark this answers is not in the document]
amandahla: Thanks!

src/charm.py
amandahla: Changed, thanks

src/charm.py
amandahla: Changed.

src/charm.py
amandahla: Removed :)

tests/integration/any_charm.py
```diff
@@ -8,9 +8,12 @@
 import pathlib
 import signal
 import subprocess
+from typing import Dict
 
-from any_charm_base import AnyCharmBase
-from ingress import IngressRequires
+from any_charm_base import AnyCharmBase  # type: ignore[import]
```
jdkandersson, line 13: I think these can be avoided by adding the dependencies to the `lint` job that `mypy` is looking for

tests/integration/any_charm.py
```diff
@@ -41,6 +44,49 @@ def update_ingress(self, ingress_config):
...
+    def _has_required_fields(self, rel: Dict) -> bool:
+        """Check for required fields in relation
+
+        Args:
+            rel: relation to check
+
+        Returns:
+            Returns true if all fields exist
+        """
+        return "service-hostname" in rel and "service-name" in rel and "service-port" in rel
```
jdkandersson, line 56: How about `return all(key in rel for key in ("service-hostname", "service-name", "service-port"))`?

tests/integration/any_charm.py
```diff
@@ -41,6 +44,49 @@ def update_ingress(self, ingress_config):
...
+        """Get ingress config from relation or default
+
+        Returns:
+            The ingress config to be used
+        """
+        if self._has_ingress_relation() and self._has_app_data():
+            rel = self.model.relations["ingress"][0].data[self.app]
+            if self._has_required_fields(rel):
+                return {
+                    "service-hostname": rel["service-hostname"],
```
jdkandersson, line 84: Would it be worth making these constants? They seem to be used in a few places

tests/integration/conftest.py
```diff
@@ -10,7 +10,7 @@
 from pathlib import Path
 from typing import List
 
-import kubernetes
+import kubernetes  # type: ignore[import]
```
jdkandersson, line 13: Similar to earlier, this could potentially be resolved if `kubernetes` is added to the `deps` of the `[testenv:lint]` environment in `tox.ini`

tests/integration/test_relation.py
```diff
@@ -43,7 +43,7 @@ async def build_and_deploy(ops_test: OpsTest, run_action):
 
     async def build_and_deploy_ingress():
         charm = await ops_test.build_charm(".")
-        return await ops_test.model.deploy(
+        return await ops_test.model.deploy(  # type: ignore[union-attr]
```
jdkandersson, line 46: How about adding an `assert ops_test.model` somewhere earlier so that `ops_test.model` is guaranteed not to be `None` in the test?

tests/unit/test_charm.py
```diff
@@ -1,13 +1,13 @@
 # Copyright 2023 Canonical Ltd.
 # See LICENSE file for licensing details.
 
-
 import unittest
+from typing import List
```
jdkandersson, line 5: Which Python version is being targeted by the charm? If it is 3.10 or later `list` can be used instead of `typing.List` which is deprecated

tests/unit/test_library_ingress.py
```diff
@@ -60,7 +61,7 @@ class CharmWithConfigDict(NginxIngressConsumerCharm):
             def __init__(self, *args):
                 super().__init__(*args, config_dict=config_dict)
 
-        self.harness = Harness(CharmWithConfigDict, meta=META)
+        self.harness = Harness(CharmWithConfigDict, meta=META)  # type: ignore[arg-type]
```
jdkandersson, line 64: Is it worth describing the problem and why it is reasonable to ignore?

jdkandersson: Also applies for the other disable rules, see here for the standard: https://github.com/canonical/is-charms-contributing-guide/blob/main/CONTRIBUTING.md#static-code-analysis

src/charm.py
```diff
@@ -227,15 +235,15 @@ def _service_name(self):
         return self._get_config_or_relation_data("service-name", "")
 
     @property
-    def _service_port(self):
+    def _service_port(self) -> Any:
```
jdkandersson, line 238: Isn't it `int`? `Any` is kind of pointless since it essentially disables typing checks, perhaps we make a follow up story to clean up the type hints?

src/charm.py
```diff
@@ -431,25 +431,36 @@ def k8s_auth(self):
 
         self._authed = True
 
+    def _core_v1_api(self) -> Any:
```
jdkandersson, line 434: Nice! Wouldn't the type be `CoreV1Api`?

src/charm.py
```diff
@@ -431,25 +431,36 @@ def k8s_auth(self):
...
+        """Use the v1 k8s API."""
+        self.k8s_auth()
+        return kubernetes.client.CoreV1Api()
+
+    def _networking_v1_api(self) -> Any:
```
jdkandersson, line 439: Wouldn't the type be `NetworkingV1Api`? I guess for the functions add/ changed in the PR it would be good to write type hints that aren't `Any`. For the rest, I can see an argument for that it should be done in a follow up PR

tests/unit/test_charm.py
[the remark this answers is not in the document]
amandahla: ISD-417 was created to address this.

tests/integration/any_charm.py
amandahla: But this file is used inside the Any charm that has this dependency. any_charm_base is not used by the charm itself.

tests/integration/conftest.py
amandahla: Without this, I get the error "error: Skipping analyzing "kubernetes": module is installed, but missing library stubs or py.typed marker  [import]". 
It seems that kubernetes doesn't have type stubs, if I understood correctly. 
We might consider using the plugin https://pypi.org/project/kubernetes-typed/ in a further PR.

tests/unit/test_library_ingress.py
amandahla: I believe to fix this one would be necessary to change CharmType as covariant = True in the Harness class

https://mypy.readthedocs.io/en/stable/generics.html#variance-of-generic-types
https://ops.readthedocs.io/en/latest/_modules/ops/testing.html#Harness

src/charm.py
amandahla: My bad, is int, thanks.
A follow up story would be great also considering the plugin to check for Kubernetes types.

src/charm.py
amandahla: Yes, it would be nice to have a closer look at this together with the plugin for Kubernetes types + mypy.

tests/integration/any_charm.py
amandahla: Nice :)

tests/integration/test_relation.py
amandahla: My first attempt didn't work out but now it's ok, thanks.

tests/unit/test_library_ingress.py
amandahla: Comment added.

src/charm.py
amandahla: ISD-422 was created to address type hints.

tests/integration/conftest.py
amandahla: ISD-422 was created to address type hints.

tests/integration/any_charm.py
jdkandersson: Oh I see, it is something that would be added in the any charm context, so it wouldn't be possible to add this as a pip dependency

jdkandersson (approved):

merkata (approved):
Looks great!

tests/integration/any_charm.py
amandahla: Exactly :) I'll mark this as resolved but this can be reviewed in the story ISD-422.

src/charm.py
amandahla: We could work on this in a new story if it's ok for you. And for this one, we continue using the Juju label.

PR #51, opened 2023-02-07
Fix typo in adding-ingress-relation.md

amandahla:
Really minor thing that I noticed while using the doc to add Ingress library to GitHub Actions Exporter charm.

Should be updated in the discourse when the workflow is enabled again.

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -71,7 +71,7 @@ async def ip_address_list(ops_test: OpsTest, app: Application):
     # Reduce the update_status frequency until the cluster is deployed
     status_message = app.units[0].workload_status_message  # type: ignore[attr-defined]
     async with ops_test.fast_forward():
-        await ops_test.model.block_until(lambda: "Ingress IP(s)" in status_message, timeout=100)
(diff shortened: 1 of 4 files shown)


arturo-seijas (dismissed):
LGTM

mthaddon (approved):

weiiwang01 (dismissed):

tests/integration/conftest.py
```diff
@@ -148,7 +150,7 @@ async def _wait_for_ingress(ingress_name):
                 for ingress in kube.list_namespaced_ingress(ops_test.model.name).items
             ],
             wait_period=5,
-            timeout=300,
+            timeout=10 * 60,
```
jdkandersson, line 153: This was switched to `10 * 60` whereas other changes it was set to `300` rather than something like `5 * 60`? The `X * 60` to me is quite nice because you can easily tell how many minutes it is without maths 😉

amandahla: I agree. x*60 is easier to ready. Changed them all to be like that and with the same value.

weiiwang01 (approved):

varshigupta12 (approved):

PR #52, opened 2023-02-07
Change default value of rewrite-target configuration

weiiwang01:
The default value `"/"` of `rewrite-target` configuration will prevent changes of `rewrite-target` from relation.

Update the default value of the `rewrite-target` configuration to be `""`.


arturo-seijas (dismissed):
LGTM

config.yaml
```diff
@@ -81,7 +81,7 @@ options:
     description: Whether requests should be written to the `rewrite-target`
     type: boolean
   rewrite-target:
-    default: "/"
+    default: ""
```
gregory-schiano, line 84: Won't it impact existing deployments ?

weiiwang01: `rewrite-target` will be `"/"` if everything is default.

```python3
    @property
    def _rewrite_target(self) -> Any:
        """Return the rewrite target from config or relation."""
        return self._get_config_or_relation_data("rewrite-target", "/")

```

jdkandersson: Should that be noted in the option description?

tests/integration/test_relation.py
```diff
@@ -227,3 +227,54 @@ def get_ingress_annotation():
...
+        method="update_ingress",
+        kwargs=json.dumps(
+            {
+                "ingress_config": {
+                    "rewrite-target": "/foo",
+                }
+            }
+        ),
+    )
+    assert isinstance(ops_test.model, Model)
```
jdkandersson, line 259: Isn't this already done earlier in the test?

tests/integration/test_relation.py
```diff
@@ -227,3 +227,54 @@ def get_ingress_annotation():
...
+    await ops_test.model.wait_for_idle(status="active")
+    await ops_test.model.block_until(
+        lambda: "nginx.ingress.kubernetes.io/rewrite-target" in get_ingress_annotation(),
+        wait_period=5,
+        timeout=300,
+    )
+
+    ingress_annotations = get_ingress_annotation()
+    assert ingress_annotations["nginx.ingress.kubernetes.io/rewrite-target"] == "/foo"
+    await run_action(
```
jdkandersson, line 269: What is the reason for needing this at the end of the test please?

config.yaml
weiiwang01: Added, thanks!

tests/integration/test_relation.py
weiiwang01: Removed, thanks!

tests/integration/test_relation.py
weiiwang01: Just to remove this option from relation data so following test cases will not be affected.

jdkandersson: So should this be done in a fixture then? That way the cleanup gets executed even if the test fails?

tests/integration/test_relation.py
weiiwang01: Removed, thanks!

tests/integration/test_relation.py
weiiwang01: Updated to use fixture, thanks!

gregory-schiano (dismissed):
LGTM

amandahla:
Should this be changed too?

https://github.com/canonical/nginx-ingress-integrator-operator/blob/main/src/charm.py#L203

tests/integration/test_relation.py
```diff
@@ -227,3 +251,37 @@ def get_ingress_annotation():
...
     )
     assert ingress_annotations["nginx.ingress.kubernetes.io/modsecurity-snippet"]
+
+
+@pytest.mark.usefixtures("build_and_deploy", "setup_new_hostname_and_port")
+@pytest.mark.parametrize(
+    "anycharm_update_ingress_config",
+    [
+        {
+            "rewrite-target": "/foo",
```
amandahla, line 261: Would be worth testing if changing back to no value, the rewrite-target default is "/" as expected?

weiiwang01: I added a new test case for the default rewrite-target value, thanks!

gregory-schiano (dismissed):
Still LGTM

tests/integration/test_relation.py
```diff
@@ -67,6 +68,29 @@ async def build_and_deploy_ingress():
...
+    )
+    await ops_test.model.wait_for_idle(status="active")
+
+    yield config
+
+    await run_action(
+        ANY_APP_NAME,
+        "rpc",
+        method="update_ingress",
+        kwargs=json.dumps({"ingress_config": {k: "" for k in config.keys()}}),
```
jdkandersson, line 89: I think the `.keys()` can be omitted since the dictionary will yield keys if used in an iteration context. Do you think calling the variable `key` instead of `k` makes the code more readable?

`kwargs=json.dumps({"ingress_config": {key: "" for key in config}})`

abv
```diff
@@ -0,0 +1 @@
+af
```
jdkandersson, line 1: What is the purpose of this file?

tests/integration/test_relation.py
weiiwang01: Yes, you are right, removed, thanks!

abv
weiiwang01: Sorry, some typo created a new file, I have removed that.

jdkandersson (approved):

tests/integration/conftest.py
```diff
@@ -152,3 +152,34 @@ async def _wait_for_ingress(ingress_name):
...
+    kubernetes.config.load_kube_config()
+    kube = kubernetes.client.NetworkingV1Api()
+    model_name = ops_test.model.name
+
+    def _get_ingress_annotation(ingress_name: str):
+        return kube.read_namespaced_ingress(
+            ingress_name, namespace=model_name
+        ).metadata.annotations
+
+    return _get_ingress_annotation
```
yanksyoon, line 170: I'm wondering if these would be better served as helper functions rather than as fixture?
Would there be a standard or guide on these? I think having them as helper functions can remove scoping and dependencies on tests. Also it could be easier to navigate code through IDE.

yanksyoon (approved):
LGTM, a minor question on styling :)

tests/integration/conftest.py
weiiwang01: The problem of converting this as a helper function is that this function needs the `ops_test` fixture. 
If I am going to declare this as a helper function, I have to pass the `ops_test` every time I invoke this function.
While as a fixture, I can just use the `ops_test` fixture.
Both way will do, it's just a personal preference.
For the IDE part, I think there are plugins for pytest that can do fixture navigation, my IDE can do that by default.

PR #53, opened 2023-02-10
Adding custom labels

gtrkiller:
This PR focuses on the creation of a custom label for both services and ingresses, to not be dependant on the `app.juju.is/created-by=` label.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -28,7 +28,7 @@ BOOLEAN_CONFIG_FIELDS = ["rewrite-enabled"]
 # Juju defines the value of this label.
 # It has the same value as the label "app.kubernetes.io/name"
 # set in the service account associated with the application.
-CREATED_BY_LABEL = "created-by"
+CREATED_BY_LABEL = "nginx-ingress-integrator.charm.juju.is/managed-by"
 REPORT_INTERVAL_COUNT = 100
 
 
--- a/src/charm.py
+++ b/src/charm.py
@@ -25,9 +25,8 @@ from ops.model import ActiveStatus, BlockedStatus, WaitingStatus
 LOGGER = logging.getLogger(__name__)
 _INGRESS_SUB_REGEX = re.compile("[^0-9a-zA-Z]")
 BOOLEAN_CONFIG_FIELDS = ["rewrite-enabled"]
-# Juju defines the value of this label.
-# It has the same value as the label "app.kubernetes.io/name"
-# set in the service account associated with the application.
+# We set this value to be unique for this deployed juju application
+# so we can use it to identify resources created by this charm
 CREATED_BY_LABEL = "nginx-ingress-integrator.charm.juju.is/managed-by"
 REPORT_INTERVAL_COUNT = 100
 
--- a/src/charm.py
+++ b/src/charm.py
@@ -20,7 +20,7 @@ from charms.nginx_ingress_integrator.v0.ingress import (  # type: ignore[import]
 )
 from ops.charm import CharmBase
 from ops.main import main
-from ops.model import ActiveStatus, BlockedStatus
+from ops.model import ActiveStatus, BlockedStatus, WaitingStatus
 
 LOGGER = logging.getLogger(__name__)
 _INGRESS_SUB_REGEX = re.compile("[^0-9a-zA-Z]")
@@ -742,7 +742,9 @@ class NginxIngressCharm(CharmBase):
         for hostname, paths in ingress_paths.items():
             # Use the metadata from the first rule.
             initial_ingress = hostname_ingress[hostname]
-            add_ingress = self._create_k8s_ingress_obj(hostname, initial_ingress, paths)
+            add_ingress = self._create_k8s_ingress_obj(
+                hostname, initial_ingress, paths, self.app.name
+            )
             ingress_objs.append(add_ingress)
 
         return ingress_objs
@@ -752,7 +754,9 @@ class NginxIngressCharm(CharmBase):
         ingress_name = _INGRESS_SUB_REGEX.sub("-", hostname)
         return f"{ingress_name}-ingress"
 
-    def _create_k8s_ingress_obj(self, svc_hostname: str, initial_ingress: Any, paths: Any) -> Any:
+    def _create_k8s_ingress_obj(
+        self, svc_hostname: str, initial_ingress: Any, paths: Any, label: str
+    ) -> Any:
         """Create a Kubernetes Ingress Resources with the given data."""
         # Create a Ingress Object with the new ingress rules and return it.
         rule = kubernetes.client.V1IngressRule(
@@ -777,6 +781,7 @@ class NginxIngressCharm(CharmBase):
             metadata=kubernetes.client.V1ObjectMeta(
                 name=ingress_name,
                 annotations=initial_ingress.metadata.annotations,
+                labels={CREATED_BY_LABEL: label},
             ),
             spec=new_spec,
         )
@@ -846,6 +851,7 @@ class NginxIngressCharm(CharmBase):
                 self._define_services()
                 self._define_ingresses()
                 msgs = []
+                self.unit.status = WaitingStatus("Waiting for ingress IP availability")
                 ingress_ips = self._report_ingress_ips()
                 if ingress_ips:
                     msgs.append(f"Ingress IP(s): {', '.join(ingress_ips)}")
--- a/src/charm.py
+++ b/src/charm.py
@@ -28,7 +28,7 @@ BOOLEAN_CONFIG_FIELDS = ["rewrite-enabled"]
 # Juju defines the value of this label.
 # It has the same value as the label "app.kubernetes.io/name"
 # set in the service account associated with the application.
-CREATED_BY_LABEL = "Ingress-label"
+CREATED_BY_LABEL = "created-by"
 REPORT_INTERVAL_COUNT = 100
 
 
@@ -280,7 +280,7 @@ class _ConfigOrRelation:
         """Return the whitelist-source-range config option."""
         return self._get_config("whitelist-source-range")
 
-    def _get_k8s_service(self, label: str) -> Any:
+    def _get_k8s_service(self, label: str) -> kubernetes.client.V1Service:
         """Get a K8s service definition.
 
         Args:
@@ -304,7 +304,7 @@ class _ConfigOrRelation:
             ),
         )
 
-    def _get_k8s_ingress(self, label: str) -> Any:
+    def _get_k8s_ingress(self, label: str) -> kubernetes.client.V1Ingress:
         """Get a K8s ingress definition.
 
         Args:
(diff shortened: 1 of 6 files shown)


src/charm.py
```diff
@@ -28,7 +28,7 @@
 # Juju defines the value of this label.
 # It has the same value as the label "app.kubernetes.io/name"
 # set in the service account associated with the application.
-CREATED_BY_LABEL = "app.juju.is/created-by="
+CREATED_BY_LABEL = "Ingress-label"
```
jdkandersson, line 31: Should the variable be renamed now that the label is no longer `created-by`? Also, minor thing, should it be `ingress-label`, so be all lower case?

jdkandersson, line 31: Also, how would this work on upgrades? E.g., when an application currently using the `app.juju.is/created-by=` label upgrades to this version of the charm, will everything still work? Should we have an integration test to confirm the upgrade?

src/charm.py
```diff
@@ -280,12 +280,18 @@ def _whitelist_source_range(self) -> Any:
         """Return the whitelist-source-range config option."""
         return self._get_config("whitelist-source-range")
 
-    def _get_k8s_service(self) -> Any:
-        """Get a K8s service definition."""
+    def _get_k8s_service(self, label: str) -> Any:
```
jdkandersson, line 283: Should the type be updated to `V1Service` given we are updating the function?

src/charm.py
```diff
@@ -298,8 +304,12 @@ def _get_k8s_service(self) -> Any:
             ),
         )
 
-    def _get_k8s_ingress(self) -> Any:
-        """Get a K8s ingress definition."""
+    def _get_k8s_ingress(self, label: str) -> Any:
```
jdkandersson, line 307: Should the return type be `V1Ingress`?

src/charm.py
```diff
@@ -655,7 +661,9 @@ def _define_ingresses(self, excluded_relation=None) -> None:  # type: ignore[no-
         # Filter out any cases in which we don't have data set (e.g.: missing relation data)
         config_or_relations = filter(self._has_required_fields, config_or_relations)
 
-        ingresses = [conf_or_rel._get_k8s_ingress() for conf_or_rel in config_or_relations]
+        ingresses = [
+            conf_or_rel._get_k8s_ingress(self.app.name) for conf_or_rel in config_or_relations
+        ]
 
         ingresses = self._process_ingresses(ingresses)
```
jdkandersson, line 668: What processing does this function do? Maybe the result should be assigned to a different variable name since what it points to has changed to now be processed?

src/charm.py
gtrkiller: The upgrade process is not yet done, that's why this PR is still a draft. I need the PR to test in the CI since it throws different results locally. Also will correct the rest of your comments after the tests start passing.

merkata (dismissed):
Looks good! I wonder if we shouldn't namespace the label like juju is doing, though that is just food for thought.

src/charm.py
jdkandersson: I don't follow, `_process_ingresses` doesn't do anything?

tests/integration/conftest.py
```diff
@@ -57,8 +57,9 @@ async def app(ops_test: OpsTest, app_name: str):
 
     # Add required relations
     await ops_test.model.add_relation(hello_kubecon_app_name, app_name)
-    await ops_test.model.wait_for_idle(timeout=10 * 60)
-
+    await ops_test.model.wait_for_idle(
+        timeout=10 * 60, status=ActiveStatus.name  # type: ignore[has-type]
```
jdkandersson, line 61: Could this go to the top of the file as a constant so we don't have to endlessly repeat the `type: ignore[has-type]`?

tests/integration/conftest.py
```diff
@@ -70,10 +71,8 @@ async def ip_address_list(ops_test: OpsTest, app: Application):
     """
     # Reduce the update_status frequency until the cluster is deployed
     async with ops_test.fast_forward():
-        status_message = app.units[0].workload_status_message  # type: ignore[attr-defined]
-        await ops_test.model.block_until(
-            lambda: "Ingress IP(s)" in status_message, timeout=15 * 60
-        )
+        await ops_test.model.wait_for_idle(status="active")
+    status_message = app.units[0].workload_status_message  # type: ignore[attr-defined]
```
jdkandersson, line 75: For the `type: ignore` this standard asks for them to be documented with a reason why it is reasonable to ignore them: https://github.com/canonical/is-charms-contributing-guide/blob/main/CONTRIBUTING.md#static-code-analysis. Is that applicable here?

jdkandersson (commented):
I still don't see an upgrade test, maybe I missed it?

jdkandersson:
Since we are making a change to the behaviour (a label is now a different value), an upgrade test from a version of the charm that used to old label to the new label would help us ensure that we don't cause any issues for any users that upgrade/ What do you think?

mthaddon:
> Looks good! I wonder if we shouldn't namespace the label like juju is doing, though that is just food for thought.

Yeah, I kind of agree that the label just being `created-by` is a big too generic. `created-by-juju-application` seems better, but looking at https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/ this one jumps out as being relevant:
```
Key | Description | Example | Type
-- | -- | -- | --
app.kubernetes.io/managed-by | The tool being used to manage the operation of an application | helm | string
-- | -- | -- | --
```
Based on this I'd go with `app.kubernetes.io/managed-by=juju-application-{self.app.name}` after confirming this label isn't already set. Would be interested in other people's thoughts about this as well.

gtrkiller:
Regarding the new and old labels, the old label (app.juju.is/created-by=) is still there, since it is created on Juju's side and not by the charm. That means we now have two labels, the juju label and our new custom label. So, is it really necessary to have an upgrade test since the old label hasn't been removed?

src/charm.py
gtrkiller: It creates the actual Ingress objects from the data defined in define_ingresses.

amandahla:
> > Looks good! I wonder if we shouldn't namespace the label like juju is doing, though that is just food for thought.
> 
> Yeah, I kind of agree that the label just being `created-by` is a big too generic. `created-by-juju-application` seems better, but looking at https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/ this one jumps out as being relevant:
> 
> ```
> Key                          | Description                                                   | Example | Type
> app.kubernetes.io/managed-by | The tool being used to manage the operation of an application | helm    | string
> ```
> 
> Based on this I'd go with `app.kubernetes.io/managed-by=juju-application-{self.app.name}` after confirming this label isn't already set. Would be interested in other people's thoughts about this as well.

I liked this idea but I think it might conflict here:
https://github.com/juju/juju/blob/develop/caas/kubernetes/provider/constants/labels.go#L19
For namespaces:
```
mymodel                      Active   4h9m    app.kubernetes.io/managed-by=juju,kubernetes.io/metadata.name=mymodel,model.juju.is/name=mymodel
```

jdkandersson:
> Regarding the new and old labels, the old label (app.juju.is/created-by=) is still there, since it is created on Juju's side and not by the charm. That means we now have two labels, the juju label and our new custom label. So, is it really necessary to have an upgrade test since the old label hasn't been removed? The ingress and service definition/update takes place only when the config or relations are changed, not in an upgrade event. I tested an upgrade and the ingress worked just fine.

I was more thinking that during an upgrade the new custom label may not be present when the charm code assumes it is there - hence the upgrade test. It would be good to see the test methodology and the outcome just to see if anyone has any feedback

mthaddon:
> I'm considering the idea of using the same label "app.juju.is/created-by". So will be applied even if is not done by Juju.

I think this is a great idea. And thanks for finding the issue with my other suggestion. I'd switch my vote to this approach.

weiiwang01:
Another approach is to create a new label prefix for the Nginx ingress integrator charm. The advantage of this option is that we can create as many labels as we want without having to worry about conflicts.

For example:
```
nginx-ingress-integrator.charm.canonical.com/managed-by=<ingress-app-name>
nginx-ingress-integrator.charm.canonical.com/created-for=<app-name>
```

amandahla:
> Another approach is to create a new label prefix for the Nginx ingress integrator charm. The advantage of this option is that we can create as many labels as we want without having to worry about conflicts.
> 
> For example:
> 
> ```
> nginx-ingress-integrator.charm.canonical.com/managed-by=<ingress-app-name>
> nginx-ingress-integrator.charm.canonical.com/created-for=<app-name>
> ```

I'm afraid that in this case would be redundant (and pretty close to the limit of [label size](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set))
Also, I guess that a valid DNS should be used like juju.is

weiiwang01:
> > Another approach is to create a new label prefix for the Nginx ingress integrator charm. The advantage of this option is that we can create as many labels as we want without having to worry about conflicts.
> > For example:
> > ```
> > nginx-ingress-integrator.charm.canonical.com/managed-by=<ingress-app-name>
> > nginx-ingress-integrator.charm.canonical.com/created-for=<app-name>
> > ```
> 
> I'm afraid that in this case would be redundant (and pretty close to the limit of [label size](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set)) Also, I guess that a valid DNS should be used like juju.is

The limit of 63 characters only applies to the name segment i.e. the segment after `/`, the prefix segment can have another 253 characters.

And yes, we can change the prefix to `nginx-ingress-integrator.charm.juju.is`, the exact prefix value is still up to debate.

weiiwang01:
> > > Another approach is to create a new label prefix for the Nginx ingress integrator charm. The advantage of this option is that we can create as many labels as we want without having to worry about conflicts.
> > > For example:
> > > ```
> > > nginx-ingress-integrator.charm.canonical.com/managed-by=<ingress-app-name>
> > > nginx-ingress-integrator.charm.canonical.com/created-for=<app-name>
> > > ```
> > 
> > 
> > I'm afraid that in this case would be redundant (and pretty close to the limit of [label size](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set)) Also, I guess that a valid DNS should be used like juju.is
> 
> The limit of 63 characters only applies to the name segment i.e. the segment after `/`, the prefix segment can have another 253 characters.
> 
> And yes, we can change the prefix to `nginx-ingress-integrator.charm.juju.is`, the exact prefix value is still up to debate.

I think if we choose the customized prefix solution here, we will create a new pattern / convention. Other charm that manages the Kubernetes resources can all have similar label names like this: `<charm-name>.charm.juju.is/<name>` if they want to add labels to the resources.

mthaddon:
There have been a number of comments about this on the discourse thread, but let's go ahead with `nginx-ingress-integrator.charm.juju.is/managed-by` for now and we can easily update this if changed later.

mthaddon (approved):
LGTM, thx

amandahla (approved):
Nice :)

PR #54, opened 2023-02-21
[DO NOT MERGE] Add the nginx-route relation and nginx_route charm library

weiiwang01:


lib/charms/nginx_ingress_integrator/v0/nginx_route.py
```diff
@@ -0,0 +1,383 @@
+# Copyright 2023 Canonical Ltd.
+# Licensed under the Apache2.0, see LICENCE file in charm source for details.
+"""Library for the nginx-route relation.
+
+This library contains the require and provide functions for handling
```
amandahla, line 5: Do you think would be relevant to add some comment here mentioning that this lib/relation should be used instead of the deprecated ingress one?

src/charm.py
```diff
@@ -413,19 +416,52 @@ def __init__(self, *args) -> None:  # type: ignore[no-untyped-def]
...
+            event: not used.
+        """
+        self._on_config_changed(event)
+        status = self.unit.status
+        connected_apps = set()
+        for relation in self.model.relations["ingress"]:
+            remote_app = relation.app
+            if remote_app is not None:
+                connected_apps.add(remote_app.name)
+        if connected_apps:
```
amandahla, line 450: Would be possible to have an ingress and a nginx-route relation at the same time?

lib/charms/nginx_ingress_integrator/v0/nginx_route.py
```diff
@@ -0,0 +1,383 @@
...
+        session_cookie_max_age: configure Nginx ingress integrator
+            session-cookie-max-age option via relation, optional.
+        tls_secret_name: configure Nginx ingress integrator
+            tls-secret-name option via relation, optional.
+        nginx_route_relation_name: Specifies the relation name of
+            the relation handled by this requirer class. The relation
+            must have the nginx-route interface.
+    """
+    config: typing.Dict[str, typing.Union[str, int, bool]] = {}
+    if service_hostname is not None:
```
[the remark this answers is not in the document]
jdkandersson: I do like the docstring and that the arguments are listed in the function. I can think of a few hacks to capture the kwargs whilst also retaining this, (e.g., involving decorators and functools.wraps), they aren't pretty tough.

One nice alternative that we could consider is uding a TypedDict because that also comes with autocomplete. Another option is to use a bunch of `typing.NamedTuple`, including aggregating similar arguments into sub tuples. That is something like what I have seen other libraries do in these cases. So you would have something like:

```python
# library code
from typing import NamedTuple

class Service(NamedTuple):
    hostname: str
    name: str
    ...

class Config(NamedTuple):
    service: Service
    ...

def require_nginx_route(..., config: Config, ...) -> None: ...

# user code
from <wherever> import Service, Config

class Charm(CharmBase):
    def __init__(self):
        service = Service(...)
        config = COnfig(service=service, ...)
        require_nginx_route(self, config, ...)
```

lib/charms/nginx_ingress_integrator/v0/nginx_route.py
```diff
@@ -0,0 +1,383 @@
...
+        config["session-cookie-max-age"] = session_cookie_max_age
+    if tls_secret_name is not None:
+        config["tls-secret-name"] = tls_secret_name
+
+    _NginxRouteRequirer(
+        charm=charm, config=config, nginx_route_relation_name=nginx_route_relation_name
+    )
+
+
+class _NginxRouteProvider(Object):
```
jdkandersson, line 255: Is it necessary to inherit from `Object`?

src/charm.py
```diff
@@ -18,6 +18,9 @@
     IngressCharmEvents,
     IngressProvides,
 )
+from charms.nginx_ingress_integrator.v0.nginx_route import (  # type: ignore[import]
```
jdkandersson, line 21: Should there be a comment here explaining the `type: ignore`?

src/charm.py
```diff
@@ -413,19 +416,52 @@ def __init__(self, *args) -> None:  # type: ignore[no-untyped-def]
...
+
+        The same functionality as _on_config_changed, but will add warning message if there are
+        applications connected via ingress relation.
+
+        Args:
+            event: not used.
+        """
+        self._on_config_changed(event)
+        status = self.unit.status
+        connected_apps = set()
```
jdkandersson, line 445: What about `connected_apps = {relation.app.name for relation in self.model.relations["ingress"] if relation.app is not None`}?

src/charm.py
```diff
@@ -413,19 +416,52 @@ def __init__(self, *args) -> None:  # type: ignore[no-untyped-def]
...
+                f"app [{connected_app_names}] connected via deprecated ingress relation, "
+                f"please update to nginx-route relation; {status.message}"
+            )
+            self.unit.status = status.from_name(status.name, warning)
+
     @property
     def _all_config_or_relations(self) -> Any:
         """Get all configuration and relation data."""
-        all_relations = self.model.relations["ingress"] or [None]  # type: ignore[list-item]
+        all_relations = self.model.relations["ingress"][:]
```
jdkandersson, line 461: What about `all_relations = list(itertools.chain(self.model.relations["ingress"][:], self.model.relations["nginx-route"]))`. I think the syntax is right, not 100% sure though

src/charm.py
```diff
@@ -413,19 +416,52 @@ def __init__(self, *args) -> None:  # type: ignore[no-untyped-def]
...
+        all_relations.extend(self.model.relations["nginx-route"])
+        if not all_relations:
+            all_relations = [None]  # type: ignore[list-item]
```
jdkandersson, line 464: This looks weird, should we fix that in this PR?

src/charm.py
```diff
@@ -435,7 +471,7 @@ def _all_config_or_relations(self) -> Any:
     @property
     def _multiple_relations(self) -> bool:
         """Return a boolean indicating if we're related to multiple applications."""
-        return len(self.model.relations["ingress"]) > 1
+        return len(self.model.relations["ingress"] + self.model.relations["nginx-route"]) > 1
```
jdkandersson, line 474: I think this creates a new list, what about `len(self.model.relations["ingress"]) + len(self.model.relations["nginx-route"])`

tests/integration/any_charm_nginx_route.py
```diff
@@ -0,0 +1,77 @@
+# Copyright 2023 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+# pylint: disable=import-error,consider-using-with,duplicate-code
```
jdkandersson, line 4: Should this document why these are reasonable to disable?

jdkandersson (commented):
I'll leave it at this for now, have not looked at the tests yet

src/charm.py
weiiwang01: From juju's perspective, it can have two relation at the same time. But if two relation provides the same information (service-name ,etc), it might cause errors in the Nginx ingress integrator charm.

lib/charms/nginx_ingress_integrator/v0/nginx_route.py
```diff
@@ -0,0 +1,383 @@
+# Copyright 2023 Canonical Ltd.
```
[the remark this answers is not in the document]
weiiwang01: I think this should be v0 since this is a new charm library not a successor of the ingress charm library.

lib/charms/nginx_ingress_integrator/v0/nginx_route.py
```diff
@@ -0,0 +1,383 @@
...
+LIBID = "c13d5d639bcd09f8c4f5b195264ed53d"
+
+# Increment this major API version when introducing breaking changes
+LIBAPI = 0
+
+# Increment this PATCH version before using `charmcraft publish-lib` or reset
+# to 0 if you are raising the major API version
+LIBPATCH = 1
+
+__all__ = ["require_nginx_route", "provide_nginx_route"]
```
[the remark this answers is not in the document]
weiiwang01: This is just a hint for what functions library users should import.

lib/charms/nginx_ingress_integrator/v0/nginx_route.py
```diff
@@ -0,0 +1,383 @@
...
+    """This class defines the functionality for the 'requires' side of the 'nginx-route' relation.
+
+    Hook events observed:
+        - relation-changed
+    """
+
+    def __init__(
+        self,
+        charm: CharmBase,
+        config: typing.Dict[str, typing.Union[str, int, bool]],
```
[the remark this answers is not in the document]
weiiwang01: This is a private class in the charm library, the input is guarded by the `require_nginx_route` function.

lib/charms/nginx_ingress_integrator/v0/nginx_route.py
```diff
@@ -0,0 +1,383 @@
...
+        super().__init__(charm, nginx_route_relation_name)
+        self._charm: CharmBase = charm
+        self._nginx_route_relation_name = nginx_route_relation_name
+        self._charm.framework.observe(
+            self._charm.on[self._nginx_route_relation_name].relation_changed,
+            self._config_reconciliation,
+        )
+        # Set default values.
+        self._config: typing.Dict[str, typing.Union[str, int, bool]] = {
+            "service-namespace": self._charm.model.name
```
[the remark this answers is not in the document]
weiiwang01: Updated, thanks!

lib/charms/nginx_ingress_integrator/v0/nginx_route.py
```diff
@@ -0,0 +1,383 @@
...
+        self._config.update(config)
+        self._config_reconciliation(None)
+
+    def _config_reconciliation(self, _event: typing.Any = None) -> None:
+        """Update the nginx-route relation data to be exactly as defined by config."""
+        if not self._charm.model.unit.is_leader():
+            return
+        for relation in self._charm.model.relations[self._nginx_route_relation_name]:
+            relation_app_data = relation.data[self._charm.app]
+            delete_keys = set(r for r in relation_app_data if r not in self._config)
```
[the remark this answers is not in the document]
weiiwang01: Updated, thanks!

lib/charms/nginx_ingress_integrator/v0/nginx_route.py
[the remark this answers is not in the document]
jdkandersson: would it be helpful to have a definition of all the environment variables that are allowed for internal consistency and documentation?

lib/charms/nginx_ingress_integrator/v0/nginx_route.py
```diff
@@ -0,0 +1,383 @@
...
+            for delete_key in delete_keys:
+                del relation_app_data[delete_key]
+            for relation_field, relation_data_value in self._config.items():
```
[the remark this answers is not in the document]
jdkandersson: If a given operation isn't supported because `relation_app_data`, shouldn't the tests fail?

lib/charms/nginx_ingress_integrator/v0/nginx_route.py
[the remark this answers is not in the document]
weiiwang01: The reason why I use function arguments here instead of `kwargs` is that:
- function arguments have better IDE support compare to `kwargs` with TypedDict
- function arguments are checked by Python, while the the type hint of `kwargs` is not enforced during the runtime
- TypedDict's support for omitting key requires Python 3.11

One concern about the NamedTuple approach is that users now require to import another class to use the charm library.

I think in this case the complexity of implementation (multiple `if`s) is acceptable for the simplicity of the interface.

lib/charms/nginx_ingress_integrator/v0/nginx_route.py
weiiwang01: This is the `Object` imported from the `ops.framework`.

src/charm.py
weiiwang01: I just removed all `# type: ignore[import]`, it will not be necessary after I add the mypy configuration to `pyproject.toml`. Thanks!

src/charm.py
weiiwang01: Updated, thanks!

lib/charms/nginx_ingress_integrator/v0/nginx_route.py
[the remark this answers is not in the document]
jdkandersson: I think you are right that on balance, the arguments are a better way of doing things in this case

Just a note, for `TypedDict`, it has been possible I believe since their introduction to declare that not all keys have been required using the `total` argument. I have written a blog post on this topic: https://jdkandersson.com/2020/01/27/python-typeddict-arbitrary-key-names-with-totality/

src/charm.py
weiiwang01: I think it might be too involved to fix this problem in the same PR. We can create a new PR for this problem.

src/charm.py
weiiwang01: Fixed, thanks!

lib/charms/nginx_ingress_integrator/v0/nginx_route.py
jdkandersson: Hmm, do you know what that class does? It is quite confusing given the `object` inbuilt type. Maybe we can give the `ops` team feedback on potentially improving the name so that it is clearer what that class does?

src/charm.py
jdkandersson: Ok, sounds good

lib/charms/nginx_ingress_integrator/v0/nginx_route.py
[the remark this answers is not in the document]
weiiwang01: The problem for the Nginx ingress integrator charm configuration dictionary is that the `service-hostname`, `service-name`, and `service-port` are required, while other fields are not. If using the `total` argument in the `TypedDict`, all fields will be optional.

lib/charms/nginx_ingress_integrator/v0/nginx_route.py
weiiwang01: It's the base class for all entities in the `ops.framework`. And yes, it's confusion.

tests/integration/any_charm_nginx_route.py
weiiwang01: Added, thanks!

lib/charms/nginx_ingress_integrator/v0/nginx_route.py
```diff
@@ -0,0 +1,383 @@
...
+                relation_app_data[relation_field] = str(relation_data_value)
+
+
+def require_nginx_route(  # pylint: disable=too-many-locals,too-many-branches
```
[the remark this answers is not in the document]
weiiwang01: See the comments below for using `kwargs`.

lib/charms/nginx_ingress_integrator/v0/nginx_route.py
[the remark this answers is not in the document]
weiiwang01: Let me check, thanks!

lib/charms/nginx_ingress_integrator/v0/nginx_route.py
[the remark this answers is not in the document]
weiiwang01: In that case, this document will be a duplicate of the document in the `require_nginx_route`. Maintaining two separated documents is prone to error.

lib/charms/nginx_ingress_integrator/v0/nginx_route.py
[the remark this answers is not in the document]
jdkandersson: It is possible to use a class hierarchy where some have the `total=True` and others `total=False`, the linked blog post above gives an example

lib/charms/nginx_ingress_integrator/v0/nginx_route.py
jdkandersson: @benhoyt Any thoughts?

lib/charms/nginx_ingress_integrator/v0/nginx_route.py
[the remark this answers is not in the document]
weiiwang01: The relation data class has the `update` method. Updated, thanks!

src/charm.py
weiiwang01: I think the `+` operator alone will create a new list. 
Updated to this:
```
        all_relations = self.model.relations["ingress"] + self.model.relations["nginx-route"]
```

src/charm.py
amandahla: Do you think is worth checking in this case?

weiiwang01: The case where an application has both ingress and nginx-route relations is no different than the case of having two ingress relations or two nginx-route relations. Having multiple relations is not an issue. The problem arises when two or more relations provide the same data, such as `service-hostname` or `service-name`. There are already checks in place to prevent duplicate data, so I don't think additional checks are necessary.

lib/charms/nginx_ingress_integrator/v0/nginx_route.py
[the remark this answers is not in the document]
weiiwang01: Learned, very nice solution 👍

lib/charms/nginx_ingress_integrator/v0/nginx_route.py
```diff
@@ -0,0 +1,386 @@
...
+            return
+
+        relation_name = event.relation.name
+        remote_app = event.app
+        if remote_app is None:
+            raise RuntimeError("_on_relation_changed was triggered by a broken relation.")
+
+        if not event.relation.data[remote_app]:
+            logger.info(
+                "%s hasn't finished configuring, waiting until relation is changed again.",
```
arturo-seijas, line 312: I find this a little vague. Can we rephrase it to something like "waiting until the relation data is populated"?

metadata.yaml
```diff
@@ -14,3 +14,6 @@ assumes:
 provides:
   ingress:
     interface: ingress
+
```
arturo-seijas, line 17: Picky comment, I'd remove this line

arturo-seijas (commented):
Just some minor comments

lib/charms/nginx_ingress_integrator/v0/nginx_route.py
benhoyt: Yeah, it's probably not a great name choice in `ops`. If you `from ops import framework` then you'd refer to it as `framework.Object`, which would be a bit clearer. In future we're [planning](https://github.com/canonical/operator/issues/731) to make it so all public names are accessible via just `import ops` so you'll be able to say `ops.Object`, but that's not available yet.

In any case, I believe you only need to inherit from `ops.framework.Object` if you're a "framework object" like an event that's going to need to be serialized (every `Object` has a unique path). In this case I don't think you need that -- it looks like the type *has* events but *is not* an event. (I haven't take a super-close look though.)

jdkandersson: Makes sense, `framework.Object` if it is needed or remove it otherwise. Perhaps worth updating the documentation of the class to explain when it is necessary? It isn't immediately obvious, at least to me, from the docstring:

```Python
class Object:
    """Initialize an Object as a new leaf in :class:`Framework`, identified by `key`.

    Args:
        parent: parent node in the tree.
        key: unique identifier for this object.

    Every object belongs to exactly one framework.

    Every object has a parent, which might be a framework.

    We track a "path to object," which is the path to the parent, plus the object's unique
    identifier. Event handlers use this identity to track the destination of their events, and the
    Framework uses this id to track persisted state between event executions.

    The Framework should raise an error if it ever detects that two objects with the same id have
    been created.

    """
```

lib/charms/nginx_ingress_integrator/v0/nginx_route.py
weiiwang01: Updated, thanks!

metadata.yaml
weiiwang01: I removed all empty lines in the `metadata.yaml`, thanks!

lib/charms/nginx_ingress_integrator/v0/nginx_route.py
weiiwang01: I will just use import to replace all import-from statements for now.

amandahla (approved):
LGTM

tox.ini
```diff
@@ -61,7 +61,7 @@ commands =
     # codespell {[vars]lib_path}
     codespell {toxinidir} --skip {toxinidir}/.git --skip {toxinidir}/.tox \
       --skip {toxinidir}/build --skip {toxinidir}/venv \
-      --skip {toxinidir}/.mypy_cache --skip {toxinidir}/icon.svg
+      --skip {toxinidir}/.mypy_cache --skip {toxinidir}/icon.svg --skip {toxinidir}/*.log
```
arturo-seijas, line 64: You shouldn't need this anymore, as per https://github.com/canonical/operator-workflows/pull/105

src/charm.py
```diff
@@ -435,7 +468,7 @@ def _all_config_or_relations(self) -> Any:
     @property
     def _multiple_relations(self) -> bool:
         """Return a boolean indicating if we're related to multiple applications."""
-        return len(self.model.relations["ingress"]) > 1
+        return len(self.model.relations["ingress"]) + len(self.model.relations["nginx-route"]) > 1
```
arturo-seijas, line 471: Why not `if self.model.relations["ingress"] or self.model.relations["nginx-route"]`?

weiiwang01: This requires the relation number greater than 1 rather than greater or equal to 1.

arturo-seijas: Oh, right. Nvm me

tox.ini
weiiwang01: Removed, thanks!

arturo-seijas (dismissed):
LGTM

tox.ini
```diff
@@ -93,6 +93,7 @@ commands =
 description = Run static analysis tests
 deps =
     bandit
+    bandit[toml]
     toml
```
arturo-seijas, line 97: I don't think `toml` is needed anymore

weiiwang01: Removed, thanks!

arturo-seijas (approved):
LGTM

lib/charms/nginx_ingress_integrator/v0/nginx_route.py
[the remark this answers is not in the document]
jdkandersson: @weiiwang01 anything on this?

jdkandersson:
Also still looks like there are 2 failing checks

tests/integration/test_nginx_route.py
```diff
@@ -0,0 +1,192 @@
...
+@pytest_asyncio.fixture(scope="module")
+async def build_and_deploy(ops_test: OpsTest, run_action):
+    """build and deploy nginx-ingress-integrator charm.
+
+    Also deploy and relate an any-charm application for test purposes.
+
+    Returns: None.
+    """
+
+    assert isinstance(ops_test.model, Model)
```
jdkandersson, line 69: What about putting this into a fixture so this assert doesn't have to be done all over the place?

tests/integration/test_nginx_route.py
```diff
@@ -0,0 +1,192 @@
+# Copyright 2023 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+# pylint: disable=redefined-outer-name,unused-argument,duplicate-code
```
jdkandersson, line 4: Are all these needed?

tests/integration/test_nginx_route.py
```diff
@@ -0,0 +1,192 @@
...
+    """
+    arrange: given charm has been built and deployed.
+    act: access ingress IP address with correct host name in HTTP headers.
+    assert: HTTP request should be forwarded to the application.
+    """
+    response = requests_get("http://127.0.0.1/ok", host_header="any")
+
+    assert response.text == "ok"
+    assert response.status_code == 200
+    assert (
```
jdkandersson, line 105: Is this assert described in the docstring? Perhaps worth a comment/ description?

tests/integration/test_nginx_route.py
```diff
@@ -0,0 +1,192 @@
...
+    await ops_test.model.applications[ANY_APP_NAME].set_config(
+        {"src-overwrite": gen_src_overwrite(service_hostname=new_hostname)}
+    )
+    await ops_test.model.wait_for_idle(status="active")
+
+    response = requests_get("http://127.0.0.1/ok", host_header=new_hostname)
+    assert response.text == "ok"
+    assert response.status_code == 200
+
+    await ops_test.model.applications[ANY_APP_NAME].set_config(
```
jdkandersson, line 130: Is this cleanup? Should it be done in a fixture?

tests/integration/test_nginx_route.py
```diff
@@ -0,0 +1,192 @@
...
+    response = requests_get("http://127.0.0.1/ok", host_header=new_hostname)
+    assert response.text == "ok"
+    assert response.status_code == 200
+    assert "additional-hostnames" in await get_relation_data()
+
+    await ops_test.model.applications[ANY_APP_NAME].set_config(
+        {"src-overwrite": gen_src_overwrite()}
+    )
+
+    await ops_test.model.wait_for_idle(status="active")
```
jdkandersson, line 167: Similar to above, should this be done in a fixture? It looks like cleanup after the test has run

tests/integration/test_relation.py
```diff
@@ -184,12 +184,12 @@ async def test_ingress_connectivity():
     act: access ingress IP address with correct host name in HTTP headers.
     assert: HTTP request should be forwarded to the application.
     """
-    response = requests.get("http://127.0.0.1/ok", headers={"Host": "any"}, timeout=5 * 60)
```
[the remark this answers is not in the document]
weiiwang01: This is due to a bandit check error, the form of `5 * 60` will be mistaken as no timeout in bandit. I just changed to this instead.

tests/integration/any_charm_nginx_route.py
```diff
@@ -0,0 +1,78 @@
...
+
+        Args:
+            port: The port where the server is connected.
+
+        Returns:
+            The port where the server is connected.
+        """
+        www_dir = pathlib.Path("/tmp/www")
+        www_dir.mkdir(exist_ok=True)
+        ok_file = www_dir / "ok"
```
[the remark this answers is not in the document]
weiiwang01: Since the `/tmp/www` directory does not exist, it will fail.

tests/integration/test_nginx_route.py
```diff
@@ -0,0 +1,192 @@
...
+        "nginx_route.py": nginx_route_lib,
+        "nginx_route_config.json": json.dumps(nginx_route_config),
+    }
+    return json.dumps(any_charm_src_overwrite)
+
+
+def requests_get(url: str, host_header: str, retry_timeout: int = 120) -> requests.Response:
+    """Requests get, but will retry when the response status code is not 200."""
+    time_start = time.time()
+    while True:
```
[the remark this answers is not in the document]
weiiwang01: Added, thanks!

tests/integration/test_nginx_route.py
weiiwang01: I can't put this in a fixture since I have to do some tests after the reset of configuration.

tests/integration/test_nginx_route.py
weiiwang01: I just refactored the code so it wouldn't be needed any more.

tests/integration/test_nginx_route.py
weiiwang01: Same as the reply above.

tests/integration/any_charm_nginx_route.py
```diff
@@ -0,0 +1,78 @@
...
+        require_nginx_route(charm=self, **self.nginx_route_config())
+
+    @staticmethod
+    def nginx_route_config() -> Dict:
+        """Get the nginx-route configuration from a JSON file on disk.
+
+        Returns:
+            The nginx-route config to be used
+        """
+        src_path = pathlib.Path(os.path.abspath(os.path.split(__file__)[0]))
```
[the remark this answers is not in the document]
weiiwang01: Updated, thanks!

lib/charms/nginx_ingress_integrator/v0/nginx_route.py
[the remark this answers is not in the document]
jdkandersson: Actually, this doesn't look like it is code related to this charm since it is in the lib

tests/integration/test_relation.py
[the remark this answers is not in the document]
jdkandersson: Perhaps we should report that to bandit as a bug? I understand why the change now

tests/integration/any_charm_nginx_route.py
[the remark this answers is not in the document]
jdkandersson: so line 61 and 62 can stay the same, this is just for this and the next line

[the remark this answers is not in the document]
jdkandersson: There is an oppoertunity for the walrus operator, not sure which version of Python needs to be supported: `(www_dir := pathlib.Path("/tmp/www")).mkdir(exist_ok=True)`

tests/integration/test_nginx_route.py
jdkandersson: Which tests? It looks like this is the last part of the test? If the fixture is function scoped the cleanup should get executed before the next tests

weiiwang01: I thinks in the `test_update_additional_hosts` there are some test following the removal of the configuration field to test if corresponding field is removed in the relation data.

tests/integration/any_charm_nginx_route.py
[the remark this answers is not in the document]
weiiwang01: Okay, I merged Line 63 and 64. But I will keep the Line 61-52. I think the reduction of just one line is not worth the cost of readability.

tests/integration/test_nginx_route.py
jdkandersson: That indicates that the tests could be relying on execution order here? In that case it would be especially good to move the setup and teardown to a fixture so that a test could be run in isolation

weiiwang01: I don't think so, the configuration is reset in the same test case.

tox.ini
```diff
@@ -93,7 +93,7 @@ commands =
 description = Run static analysis tests
 deps =
     bandit
```
arturo-seijas, line 95: `bandit`is not needed either

weiiwang01: Removed, thanks!

jdkandersson (approved):

arturo-seijas (approved):
LGTM

lib/charms/nginx_ingress_integrator/v0/nginx_route.py
```diff
@@ -0,0 +1,386 @@
...
+        config["session-cookie-max-age"] = session_cookie_max_age
+    if tls_secret_name is not None:
+        config["tls-secret-name"] = tls_secret_name
+
+    _NginxRouteRequirer(
+        charm=charm, config=config, nginx_route_relation_name=nginx_route_relation_name
+    )
+
+
+class _NginxRouteProvider(ops.framework.Object):
```
gregory-schiano, line 258: I'm wondering if we shouldn't split requirer and provider in 2 different files like the data platform team is doing ?
See: https://github.com/canonical/data-platform-libs/tree/main/lib/charms/data_platform_libs/v0

src/charm.py
```diff
@@ -413,19 +414,51 @@ def __init__(self, *args) -> None:  # type: ignore[no-untyped-def]
...
+        applications connected via ingress relation.
+
+        Args:
+            event: not used.
+        """
+        self._on_config_changed(event)
+        status = self.unit.status
+        connected_apps = {
+            relation.app.name
+            for relation in self.model.relations["ingress"]
```
gregory-schiano, line 445: Hm wouldn't this show the warning also for apps integrated with the correct library ?

gregory-schiano (commented):
I'm stopping there, still need to review the tests

lib/charms/nginx_ingress_integrator/v0/nginx_route.py
```diff
@@ -0,0 +1,386 @@
...
+    max_body_size: typing.Optional[int] = None,
+    owasp_modsecurity_crs: typing.Optional[str] = None,
+    owasp_modsecurity_custom_rules: typing.Optional[str] = None,
+    path_routes: typing.Optional[str] = None,
+    retry_errors: typing.Optional[str] = None,
+    rewrite_target: typing.Optional[str] = None,
+    rewrite_enabled: typing.Optional[bool] = None,
+    service_namespace: typing.Optional[str] = None,
+    session_cookie_max_age: typing.Optional[int] = None,
+    tls_secret_name: typing.Optional[str] = None,
```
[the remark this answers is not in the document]
jdkandersson: We discussed something similar in this comment thread: https://github.com/canonical/nginx-ingress-integrator-operator/pull/54#discussion_r1127283060

lib/charms/nginx_ingress_integrator/v0/nginx_route.py
```diff
@@ -0,0 +1,386 @@
...
+        self._charm.framework.observe(
+            self._charm.on[self._nginx_route_relation_name].relation_changed,
+            self._config_reconciliation,
+        )
+        # Set default values.
+        self._config: typing.Dict[str, typing.Union[str, int, bool]] = {
+            "service-namespace": self._charm.model.name,
+            **config,
+        }
+        self._config_reconciliation(None)
```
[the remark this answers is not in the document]
weiiwang01: This required for the requirer side to change the relation data after the relation data initialized.

lib/charms/nginx_ingress_integrator/v0/nginx_route.py
```diff
@@ -0,0 +1,386 @@
...
+                relation_field
+                for relation_field in relation_app_data
+                if relation_field not in self._config
+            }
+            for delete_key in delete_keys:
+                del relation_app_data[delete_key]
+            relation_app_data.update({k: str(v) for k, v in self._config.items()})
+
+
+def require_nginx_route(  # pylint: disable=too-many-locals,too-many-branches
```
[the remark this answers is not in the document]
weiiwang01: I think the `require_nginx_route` function API is more intuitive and simple than the Provider class API. So I just hide the Provider class and encourage use to use the `require_nginx_route` function API.

lib/charms/nginx_ingress_integrator/v0/nginx_route.py
weiiwang01: Maybe, but I think splitting into two different libraries or keeping in the same library are just personal preference. The original `ingress` library keeps everything in the same library, so I am doing it here.

src/charm.py
weiiwang01: The correct library (`nginx-route`) will only connect with the `nginx-route` relation. So the warning message will not appear with applications using the `nginx-route` charm library.

PR #55, opened 2023-02-27
feat(ISD-482): Add proxy-read-timeout configuration

amandahla:
An application might have long-running operations that need more time to finish. The default proxy_read_timeout is 60s so if you need more than that and don't increase it, NGINX will give a 504 (Gateway Timeout) error.

This PR adds a new configuration to set proxy-read-timeout so it can be changed as required.

proxy-read-timeout: Defines a timeout in seconds for reading a response from the proxied server.

Reference [here](https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#custom-timeouts).

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -67,6 +67,10 @@ options:
       Comma separated list of the routes under the hostname that you wish to map to the relation.
       Example: "/admin,/portal" will map example.test/admin and example.test/portal only.
     type: string
+  proxy-read-timeout:
+    default: 60
+    description: Defines a timeout in seconds for reading a response from the proxied server.
+    type: int
   retry-errors:
     default: ""
     description: |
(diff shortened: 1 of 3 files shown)


jdkandersson:
> Add new configuration to set proxy-read-timeout.
> 
> proxy-read-timeout: Defines a timeout in seconds for reading a response from the proxied server.
> 
> Reference [here](https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#custom-timeouts).

Would it be worth adding the reason for needing this to the description of the PR?

arturo-seijas (approved):
LGTM

nrobinaubertin (approved):

PR #57, opened 2023-03-01
Fix type hints

arturo-seijas:


weiiwang01:
Since we are going to deprecate this charm library very soon, may I ask if there's any special reasons for this change?

arturo-seijas:
> Since we are going to deprecate this charm library very soon, may I ask if there's any special reasons for this change?

I just noticed it wasn't being linted and decided con configure tox to do so

weiiwang01 (approved):

gtrkiller (approved):
LGTM

PR #58, opened 2023-03-02
Fix the publish charm workflow

weiiwang01:
The integration test workflow from operator-workflows run multiple test modules in the same model. 

This behavior causes problems in the `test_delete_unused_ingresses` test case, since there will be some ingress resource residues from previous test runs.


arturo-seijas (approved):
LGTM

amandahla (approved):
LGTM

PR #60, opened 2023-03-22
Update the LIBID in nginx_route charm library

weiiwang01:
The current `LIBID` in the `nginx_route` charm library is incorrect and it causes the publish-lib to fail.


arturo-seijas (approved):
LGTM

mthaddon (approved):

merkata (approved):

Issue #59, opened 2023-03-13
Add Backend Protocol Config Option


kelkawi-a:
### Enhancement Proposal

Hi,

It would be useful to have a config option to set the [backend protocol](https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#backend-protocol) (e.g. to GRPC) on each resource created by the charm. A charm could be related to multiple other charms, so should offer the ability of setting different backend protocols for each relation.

weiiwang01:
Thank you for your suggestion! This seems to be not very difficult to implement.

But I just have one concern: will this option conflict with other options in the Nginx ingress integrator charm? For example, `tls-secret-name` or `owasp-modsecurity-crs`.

kelkawi-a:
I'm not well-versed in the different annotations for the ingress resource so I'm not sure. Is there a reason to believe they might conflict? Currently I am manually editing the ingress resource to include `nginx.ingress.kubernetes.io/backend-protocol: GRPC`.

weiiwang01:
For example, if the `owasp-modsecurity-crs` charm configuration is set to be `true`. The `nginx.ingress.kubernetes.io/enable-modsecurity: true` and `nginx.ingress.kubernetes.io/enable-owasp-modsecurity-crs: true` annotations will be added to the ingress resources. Could the `nginx.ingress.kubernetes.io/enable-modsecurity:true` annotation coexists with `nginx.ingress.kubernetes.io/backend-protocol: GRPC` on the same ingress resources?

weiiwang01:
I have added this new feature to our roadmap, thank you for your suggestion!

amandahla:
Fixed by https://github.com/canonical/nginx-ingress-integrator-operator/pull/66

PR #66, opened 2023-04-17
feat(ISD-616): Add backend-protocol configuration (#59)

amandahla:
Add backend-protocol  configuration.
Default: HTTP
Valid values: "HTTP", "HTTPS", "GRPC", "GRPCS", "AJP", "FCGI"
Reference: [Ingress User Guide](https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#backend-protocol ).

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -61,6 +61,7 @@ class ConflictingRoutesError(Exception):
 class InvalidBackendProtocolError(Exception):
     """Custom error that indicates invalid backend protocol."""
 
+
 class InvalidHostnameError(Exception):
     """Custom error that indicates invalid hostnames."""
 
@@ -156,9 +157,7 @@ class _ConfigOrRelation:
     @property
     def _backend_protocol(self) -> str:
         """Return the backend-protocol to use for k8s ingress."""
-        return self._get_config_or_relation_data(
-            "backend-protocol", "HTTP"
-        ).upper()
+        return self._get_config_or_relation_data("backend-protocol", "HTTP").upper()
 
     @property
     def _k8s_service_name(self) -> str:
@@ -770,8 +769,9 @@ class NginxIngressCharm(CharmBase):
                 if not invalid_hostname_check(rule.host):
                     raise InvalidHostnameError()
 
-                backend_protocol = ingress.metadata.annotations["nginx.ingress.kubernetes.io/backend-protocol"]
-                if not is_backend_protocol_valid(backend_protocol):
+                if not is_backend_protocol_valid(
+                    ingress.metadata.annotations["nginx.ingress.kubernetes.io/backend-protocol"]
+                ):
                     raise InvalidBackendProtocolError()
 
                 if rule.host not in ingress_paths:
--- a/src/charm.py
+++ b/src/charm.py
@@ -765,14 +765,14 @@ class NginxIngressCharm(CharmBase):
         for ingress in ingresses:
             # A relation could have defined additional-hostnames, so there could be more than
             # one rule. Those hostnames might also be used in other relations.
+            if not is_backend_protocol_valid(
+                ingress.metadata.annotations["nginx.ingress.kubernetes.io/backend-protocol"]
+            ):
+                raise InvalidBackendProtocolError()
+
             for rule in ingress.spec.rules:
                 if not invalid_hostname_check(rule.host):
                     raise InvalidHostnameError()
-
-                if not is_backend_protocol_valid(
-                    ingress.metadata.annotations["nginx.ingress.kubernetes.io/backend-protocol"]
-                ):
-                    raise InvalidBackendProtocolError()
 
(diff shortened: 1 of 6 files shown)


src/charm.py
```diff
@@ -147,6 +147,20 @@ def _additional_hostnames(self) -> Generator[str, None, None]:
...
+        backend_protocol_default = "HTTP"
+        backend_protocol = self._get_config_or_relation_data(
+            "backend-protocol", backend_protocol_default
+        )
+        # Disabled to reference the documentation
+        # See https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#backend-protocol # noqa: E501 pylint: disable=line-too-long
+        accepted_values = ["HTTP", "HTTPS", "GRPC", "GRPCS", "AJP", "FCGI"]
+        if backend_protocol.upper() not in accepted_values:
+            return backend_protocol_default
+        return f"{backend_protocol}".upper()
```
yanksyoon, line 162: I think this might not need an f-string :D

src/charm.py
```diff
@@ -147,6 +147,20 @@ def _additional_hostnames(self) -> Generator[str, None, None]:
...
+    @property
+    def _backend_protocol(self) -> str:
+        """Return the backend-protocol to use for k8s ingress."""
+        backend_protocol_default = "HTTP"
+        backend_protocol = self._get_config_or_relation_data(
+            "backend-protocol", backend_protocol_default
+        )
+        # Disabled to reference the documentation
+        # See https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#backend-protocol # noqa: E501 pylint: disable=line-too-long
+        accepted_values = ["HTTP", "HTTPS", "GRPC", "GRPCS", "AJP", "FCGI"]
```
yanksyoon, line 159: I think a tuple can be used here :)

yanksyoon (approved):
LGTM!, I believe @weiiwang01 was doing some testing on it too 👍 , I'm wondering if integration testing would help verify all the protocols work. Thank you for this!

src/charm.py
amandahla: Replaced :)

src/charm.py
amandahla: Replaced :)

src/charm.py
```diff
@@ -147,6 +147,20 @@ def _additional_hostnames(self) -> Generator[str, None, None]:
...
+    def _backend_protocol(self) -> str:
+        """Return the backend-protocol to use for k8s ingress."""
+        backend_protocol_default = "HTTP"
+        backend_protocol = self._get_config_or_relation_data(
+            "backend-protocol", backend_protocol_default
+        ).upper()
+        # Disabled to reference the documentation
+        # See https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#backend-protocol # noqa: E501 pylint: disable=line-too-long
+        accepted_values = ("HTTP", "HTTPS", "GRPC", "GRPCS", "AJP", "FCGI")
+        if backend_protocol not in accepted_values:
```
arturo-seijas, line 160: Perhaps it would be a better UX to block the charm if the value is invalid

src/charm.py
```diff
@@ -2,7 +2,7 @@
 # Copyright 2023 Canonical Ltd.
 # See LICENSE file for licensing details.
 
-# pylint: disable=protected-access,too-few-public-methods
+# pylint: disable=protected-access,too-few-public-methods,too-many-lines
```
gregory-schiano, line 5: Isn't there any quick win to remove this "too-many-line" ?
Any helpers that could be moved to another module ?

gregory-schiano (commented):
Isn't there any test where you test non-default backend-protocol value ?

src/charm.py
amandahla: My suggestion is to address this issue in a different story so it would be possible to consider other charms as well:
- Indico
https://github.com/canonical/indico-operator/blob/main/src/charm.py#L6
- WordPress
https://github.com/canonical/wordpress-k8s-operator/blob/main/pyproject.toml#L56

src/charm.py
amandahla: I was in doubt about that because "retry-errors" configuration option does not block so I followed the pattern.
From the config.yaml : "Unrecognised values will be ignored. The nginx default will be used if this config option is set to an empty value."
I could block or updated the doc warning about this behavior as retry-errors does. WDYT?

src/charm.py
gregory-schiano: Ok, as long as we keep track of them. I suppose also that at some point we'll put an epic to migrate towards the reference architecture.

yanksyoon (dismissed):
LGTM!

arturo-seijas (dismissed):

yanksyoon (approved):
LGTM!

PR #75, opened 2023-06-09
openssl command in docs should be lower case

plars:
minor doc fix I noticed while going through this


arturo-seijas (approved):
Thanks for your contribution

weiiwang01 (approved):

PR #85, opened 2023-08-08
Handle the start event

mthaddon:
We need to set `ActiveStatus` in the `start` event because this is a workload-less charm, so there's no `pebble-ready` event to react to. This means this is the only event (outside of `update-status`) that is fired if a pod is restarted by the k8s cluster. If we don't do anything here the charm would remain in maintenance status.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -520,7 +520,11 @@ class NginxIngressCharm(CharmBase):
             self.unit.status = status.from_name(status.name, warning)
 
     def _on_start(self, event: Any) -> None:
-        """Handle the start event."""
+        """Handle the start event.
+
+        Args:
+            event: not used.
+        """
         # We need to set ActiveStatus here because this is a workload-less
         # charm, so there's no pebble-ready event to react to. This means this
         # is the only event (outside of update-status) that is fired if a pod is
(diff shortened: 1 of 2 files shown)


src/charm.py
```diff
@@ -518,6 +519,19 @@ def _on_config_changed_with_warning(self, event: Any) -> None:
             )
             self.unit.status = status.from_name(status.name, warning)
 
+    def _on_start(self, event: Any) -> None:
```
arturo-seijas, line 522: It'd probably be better to use an underscore instead of `event` and don't document the arg in the docstring

mthaddon: Good point, updated.

src/charm.py
```diff
@@ -518,6 +519,15 @@ def _on_config_changed_with_warning(self, event: Any) -> None:
             )
             self.unit.status = status.from_name(status.name, warning)
 
+    def _on_start(self, _: Any) -> None:
```
arturo-seijas, line 522: S/Any/StartEvent

mthaddon: Have updated.

arturo-seijas (approved):
LGTM

weiiwang01 (approved):

PR #86, opened 2023-08-09
Support zero downtime upgrade from legacy ingress relation to nginx-route relation

weiiwang01:
This PR introduces the ability for mission-critical applications to seamlessly upgrade from the legacy `ingress` relation to the `nginx-route` relation without experiencing any downtime.

For applications opting to use this zero downtime transition:

1. The application should first prepare an intermediate revision that supports both `ingress` and `nginx-route` relations.
2. Upgrade the application charm to this intermediate revision.
3. Establish the `nginx-route` relation with the same relation data to `nginx-ingress-relation` while retaining the legacy `ingress` relation.
4. Ensure that all operations and processes are idle.
5. Safely remove the legacy `ingress` relation.
6. Refresh the application charm to the revision that exclusively supports `nginx-route`.

By following this sequence, applications can ensure a smooth transition with minimal risk and zero downtime.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -573,6 +573,10 @@ class NginxIngressCharm(CharmBase):
         dedup_ingress_relations = []
         for relation in ingress_relations:
             if key(relation) in nginx_route_relation_keys:
+                LOGGER.warning(
+                    "legacy ingress relation from app %s is shadowed by nginx-route relation",
+                    relation.app,
+                )
                 continue
             dedup_ingress_relations.append(relation)
         return nginx_route_relations + dedup_ingress_relations
--- a/src/charm.py
+++ b/src/charm.py
@@ -9,9 +9,11 @@
 import logging
 import re
 import time
+import typing
 from typing import Any, Dict, Generator, List, Optional, Union
 
 import kubernetes.client
+import ops
 from charms.nginx_ingress_integrator.v0.ingress import (
     RELATION_INTERFACES_MAPPINGS,
     REQUIRED_INGRESS_RELATION_FIELDS,
@@ -528,6 +530,53 @@ class NginxIngressCharm(CharmBase):
         # would remain in maintenance status.
         self.unit.status = ActiveStatus()
 
+    def _deduped_relations(self) -> List[ops.Relation]:
+        """Return a relation list with duplicates removed.
+
+        Relations are considered duplicated if they meet the following criteria:
+        1. Two are connected via the legacy ingress and the nginx-route relation endpoint
+        2. Two share the same 4-tuple (service-hostname, service-name, service-model, service-port)
+        3. Both relations have the same remote application.
+
+        In the case of duplicates, the relation connected via legacy ingress is removed.
+
+        Returns:
+            A relation list with duplicates removed.
+        """
+
+        def key(rel: Relation) -> typing.Tuple[str, ...]:
+            """Generate a key from the given relation to detect duplicates.
+
+            Args:
+                rel: the given relation object.
+
+            Returns: The key as a tuple of strings.
+
+            Raises:
+                RuntimeError: if the remote application is unknown.
+            """
+            app = rel.app
+            if not app:
+                raise RuntimeError(f"can't retrieve remote application from relation {rel}")
+            data = rel.data[app]
+            return (
+                data.get("service-hostname"),
+                data.get("service-name"),
+                data.get("service-model"),
+                data.get("service-port"),
+                app.name,
+            )
+
+        nginx_route_relations = self.model.relations["nginx-route"]
+        ingress_relations = self.model.relations["ingress"]
+        nginx_route_relation_keys = set(key(r) for r in nginx_route_relations)
+        dedup_ingress_relations = []
+        for relation in ingress_relations:
+            if key(relation) in nginx_route_relation_keys:
+                continue
+            dedup_ingress_relations.append(relation)
+        return nginx_route_relations + dedup_ingress_relations
+
     @property
     def _all_config_or_relations(self) -> Any:
         """Get all configuration and relation data.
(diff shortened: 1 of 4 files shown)


src/charm.py
```diff
@@ -528,14 +530,61 @@ def _on_start(self, _: StartEvent) -> None:
...
+                data.get("service-model"),
+                data.get("service-port"),
+                app.name,
+            )
+
+        nginx_route_relations = self.model.relations["nginx-route"]
+        ingress_relations = self.model.relations["ingress"]
+        nginx_route_relation_keys = set(key(r) for r in nginx_route_relations)
+        dedup_ingress_relations = []
+        for relation in ingress_relations:
```
arturo-seijas, line 574: Have you considered using a set instead?

weiiwang01: I try to avoid using Python objects as keys in dictionaries or sets because their hashing and equality can be not very straightforward.

src/charm.py
```diff
@@ -528,14 +530,61 @@ def _on_start(self, _: StartEvent) -> None:
...
+        2. Two share the same 4-tuple (service-hostname, service-name, service-model, service-port)
+        3. Both relations have the same remote application.
+
+        In the case of duplicates, the relation connected via legacy ingress is removed.
+
+        Returns:
+            A relation list with duplicates removed.
+        """
+
+        def key(rel: Relation) -> typing.Tuple[str, ...]:
```
jdkandersson, line 547: What is the reason for making this a nested function please?

src/charm.py
jdkandersson, line 574: list comprehension?

src/charm.py
```diff
@@ -528,14 +530,61 @@ def _on_start(self, _: StartEvent) -> None:
...
+
+        nginx_route_relations = self.model.relations["nginx-route"]
+        ingress_relations = self.model.relations["ingress"]
+        nginx_route_relation_keys = set(key(r) for r in nginx_route_relations)
+        dedup_ingress_relations = []
+        for relation in ingress_relations:
+            if key(relation) in nginx_route_relation_keys:
+                continue
+            dedup_ingress_relations.append(relation)
+        return nginx_route_relations + dedup_ingress_relations
```
jdkandersson, line 578: perhaps make this a generator?

src/charm.py
```diff
@@ -528,14 +530,61 @@ def _on_start(self, _: StartEvent) -> None:
...
+                app.name,
+            )
+
+        nginx_route_relations = self.model.relations["nginx-route"]
+        ingress_relations = self.model.relations["ingress"]
+        nginx_route_relation_keys = set(key(r) for r in nginx_route_relations)
+        dedup_ingress_relations = []
+        for relation in ingress_relations:
+            if key(relation) in nginx_route_relation_keys:
+                continue
```
amandahla, line 576: Do you think it could be useful to add a logger.debug warning about duplicates found?

weiiwang01: That's a great idea, added, thanks!

src/charm.py
weiiwang01: I need to log some message here, so a for loop is more convenient. Thanks!

src/charm.py
weiiwang01: The `key` function is only used in this function, and it's the internal implementation of the `_deduped_relations` function, so it's locally scoped.

src/charm.py
weiiwang01: Is there any special reason for using the generator here?

src/charm.py
jdkandersson: Locally scoped functions should be rare. There are certain cases where you need local context, like for decorators. Otherwise functions shouldn't be embedded in other functions because it makes them more complicated

src/charm.py
jdkandersson: Its because `.append` and combining lists using `+` can involve a lot of list machinery. With a generator, the function can just yield all the elements and the caller can put them in a list or tuple or whatever data structure they need once and get everything

weiiwang01: Commonly there are just one or two relation, talking about performance here is just premature optimization. Generator complicates the interface, unless using the generator can simplify the function code, or there's a real incentive for performance, I will not consider use the generator here.

src/charm.py
weiiwang01: Why locally scoped functions should be rare? The `key` function here is very simple and it reduce the human error of create the key separately for ingress relation and nginx-route relation. I don't see why it's complicating the code. A private function means it can be used by other methods in the class, which is not what we need.

src/charm.py
jdkandersson: I don't think it complicates the interface, here is what the code would be:

```python
nginx_route_relations = self.model.relations["nginx-route"]
yield from nginx_route_relations
ingress_relations = self.model.relations["ingress"]
nginx_route_relation_keys = set(key(r) for r in nginx_route_relations)
for relation in ingress_relations:
    if key(relation) in nginx_route_relation_keys:
        LOGGER.warning(
            "legacy ingress relation from app %s is shadowed by nginx-route relation",
            relation.app,
        )
        continue
    yield relation
```

and then it can be called with `list(def _deduped_relations())`

src/charm.py
jdkandersson: Because locally scoped functions make the function it is incorporated in more complicated and that locally scoped function has access to the outer scope local variables which can lead to unintended coupling between the functions and weird bugs. A private function can be used by other methods in the class, but doesn't have to be used

weiiwang01: This `key` function just have 5 statements, and I have already considered the variable scoping when I created this function. If you have find any incorrect variable accessing, please let me know. This coupling is intended since the `key` function is an integrating part of the `_deduped_relations` function. Isn't it better if it can not even be used by other methods in the class instead of relying on conventions?

src/charm.py
weiiwang01: If the caller only want a list, why return a generator and let the caller convert that to a list?

src/charm.py
jdkandersson: could call it `_gen_rel_dedup_key`

src/charm.py
jdkandersson: Agreed to use a tuple

src/charm.py
weiiwang01: updated, thanks

jdkandersson (approved):

amandahla (approved):
LGTM

PR #87, opened 2023-08-10
Update maintainers in metadata.yaml as required by charmhub

weiiwang01:


arturo-seijas (approved):

yanksyoon (approved):
LGTM, may I know where the reference to maintainers schema in metadata.yaml is?

weiiwang01:
> LGTM, may I know where the reference to maintainers schema in metadata.yaml is?

I have no idea, the error seems to return from the charmhub, I think it need to be a http, https or mailto url.

Issue #78, opened 2023-06-20
Ability to whitelist IPs for specific paths


kian99:
### Enhancement Proposal

Hi, 

I'm trying to figure out if I'm be able to coerce the current nginx-ingress charm to provide IP whitelisting for specific paths using the `whitelist-source-range` and `path-routes` config options. With pure Kubernetes this seems possible as described [here](https://stackoverflow.com/questions/58925853/kubernetes-ingress-whitelist-ip-for-path).

**Motivation:**
This would be useful to gate certain endpoints, e.g. a `/admin` endpoint behind a VPN, preventing users from ever getting to the application layer.

**Experimentation:**
I've so far been unsuccessful in trying to achieve this. The link above suggests creating 2 ingress resources, one that includes the ip whitelist for the `/admin` endpoint and a second that is the ingress for all other endpoints and according to kubernetes [docs](https://juju.is/docs/olm/deployment-of-juju-agents) the longer path would take precedence which would all work out. 

I've tried making two relations from my application to a single nginx-ingress charm as well as deploying 2 nginx charms, in both cases ensuring that the requiring charm provides different service names and relation names as seen below.

```python
        require_nginx_route(
            charm=self,
            service_hostname=self.config["external_hostname"],
            service_name=f"{self.app.name}_all",
            service_port=8080,
            nginx_route_relation_name="ingress-route"
        )
        require_nginx_route(
            charm=self,
            service_hostname=self.config["external_hostname"],
            service_name=f"{self.app.name}_admin",
            service_port=8080,
            path_routes="/admin",
            nginx_route_relation_name="admin-route"
        )
```
The main problem seems to be [this bit](https://github.com/canonical/nginx-ingress-integrator-operator/blob/main/src/charm.py#L209) from the source where it states,
>If there are 2 or more services configured to use the same service-hostname, the controller nginx/nginx-ingress requires them to be in the same Kubernetes Ingress object. Otherwise, Ingress will be served for only one of the services. Because of this, we'll have to group all ingresses into the same Kubernetes Resource based on their requested service-hostname.

[...]

PR #91, opened 2023-08-16
Introducing standard ingress relation support

weiiwang01:
This update integrates standard `ingress` relation support into the `nginx-ingress-integrator` charm. Concurrently, it phases out the legacy `ingress` relation.

This updated `nginx-ingress-integrator` charm isn't backward compatible with earlier versions.

Due to certain constraints of the nginx ingress controller, the `nginx-ingress-integrator` only supports the v2 version of the ingress relation with the `ip` field. However, please note that changes to include the `ip` field have not been merged as of now. See the pending changes here: [canonical/traefik-k8s-operator#220](https://github.com/canonical/traefik-k8s-operator/pull/220)

**Other notable changes are:**

- The `nginx-ingress-integrator` will now only accept one relation at a time; either one `nginx-route` relation without any `ingress` relation or vice versa.
- The configuration-only mode has been removed. Relations are required for the `nginx-ingress-integrator` to create ingress.
- The default value for the `rewrite-enabled` configuration has been modified to `false`.
- The `describe-ingresses` action is removed, since the action only prints definition of the ingress, while the charm manage four different types of resources. Only provide ingress definition might not be able to provide a full picture for user. Users should use `kubectl` for debugging.
- The service IP addresses info in the charm status is remove, since the standard ingress uses headless service which doesn't have cluster IP address, this is to align the experience of `ingress` relation and `nginx-route` relation.


src/charm.py
```diff
@@ -7,472 +7,34 @@
...
 
 
 class NginxIngressCharm(CharmBase):
-    """Charm the service.
-
-    Attrs:
-        _authed: If the charm is authed or not
-        on: Ingress charm events to handle.
-    """
+    """Charm the service."""
```
jdkandersson, line 35: Could this be expanded a bit?

src/charm.py
```diff
@@ -481,134 +43,61 @@ def __init__(self, *args) -> None:  # type: ignore[no-untyped-def]
             args: Variable list of positional arguments passed to the parent constructor.
         """
         super().__init__(*args)
-        self.framework.observe(self.on.config_changed, self._on_config_changed_with_warning)
+        if not self.unit.is_leader():
```
jdkandersson, line 46: This can be moved to the start function

src/charm.py
```diff
@@ -481,134 +43,61 @@ def __init__(self, *args) -> None:  # type: ignore[no-untyped-def]
...
+            self.unit.status = WaitingStatus(
+                "follower unit is idling, "
+                "remove follower units using `juju scale-application {self.app.name} 1`"
```
jdkandersson, line 49: Perhaps change this to something along "this charm only supports a single unit, please remove the additional units"

src/charm.py
```diff
@@ -481,134 +43,61 @@ def __init__(self, *args) -> None:  # type: ignore[no-untyped-def]
...
+        self.framework.observe(self.on.start, self._on_config_changed)
 
-        # 'ingress' relation handling.
-        self.ingress = IngressProvides(self)
-        # When the 'ingress' is ready to configure, do so.
-        self.framework.observe(self.on.ingress_available, self._on_config_changed_with_warning)
-        self.framework.observe(self.on.ingress_broken, self._on_ingress_broken)
+        self._ingress_provider = IngressPerAppProvider(charm=self)
+        self.framework.observe(self._ingress_provider.on.data_provided, self._on_config_changed)
+        self.framework.observe(self._ingress_provider.on.data_removed, self._on_config_changed)
```
jdkandersson, line 61: These should be mapped to dedicated event handlers

src/charm.py
```diff
@@ -481,134 +43,61 @@ def __init__(self, *args) -> None:  # type: ignore[no-untyped-def]
...
         self.framework.observe(self.on.describe_ingresses_action, self._describe_ingresses_action)
-        self.framework.observe(self.on.start, self._on_start)
+        self.framework.observe(self.on.start, self._on_config_changed)
 
-        # 'ingress' relation handling.
-        self.ingress = IngressProvides(self)
-        # When the 'ingress' is ready to configure, do so.
-        self.framework.observe(self.on.ingress_available, self._on_config_changed_with_warning)
-        self.framework.observe(self.on.ingress_broken, self._on_ingress_broken)
+        self._ingress_provider = IngressPerAppProvider(charm=self)
```
jdkandersson, line 59: perhaps move this above to have all the observe together

src/charm.py
```diff
@@ -481,134 +43,61 @@ def __init__(self, *args) -> None:  # type: ignore[no-untyped-def]
...
+        self.framework.observe(self._ingress_provider.on.data_provided, self._on_config_changed)
+        self.framework.observe(self._ingress_provider.on.data_removed, self._on_config_changed)
 
         provide_nginx_route(
             charm=self,
             on_nginx_route_available=self._on_config_changed,
-            on_nginx_route_broken=self._on_ingress_broken,
+            on_nginx_route_broken=self._on_config_changed,
```
jdkandersson, line 66: This should be a dedicated method

src/charm.py
```diff
@@ -481,134 +43,61 @@ def __init__(self, *args) -> None:  # type: ignore[no-untyped-def]
...
-    @property
-    def _multiple_relations(self) -> bool:
-        """Check if we're related to multiple applications.
-
-        Returns:
-            if we're related to multiple applications or not.
-        """
-        return len(self._deduped_relations()) > 1
+        Raises:
+            InvalidIngressOptionError: If both "nginx-route" and "ingress" relations are present
```
jdkandersson, line 76: This should be a pre-condition similar to whether there are any non-leader units

src/charm.py
```diff
@@ -481,134 +43,61 @@ def __init__(self, *args) -> None:  # type: ignore[no-untyped-def]
...
+        if nginx_route_relations and ingress_relations:
+            raise InvalidIngressOptionError(
+                "nginx-ingress-integrator cannot establish more than one relation at a time"
+            )
+        if nginx_route_relations:
+            relation = nginx_route_relations[0]
+        elif ingress_relations:
+            relation = ingress_relations[0]
+        else:
+            return None
```
jdkandersson, line 90: This should raise an error instead of returning None

src/charm.py
```diff
@@ -481,134 +43,61 @@ def __init__(self, *args) -> None:  # type: ignore[no-untyped-def]
...
+        nginx_route_relations = self.model.relations["nginx-route"]
+        ingress_relations = self.model.relations["ingress"]
+        if nginx_route_relations and ingress_relations:
+            raise InvalidIngressOptionError(
+                "nginx-ingress-integrator cannot establish more than one relation at a time"
+            )
+        if nginx_route_relations:
+            relation = nginx_route_relations[0]
+        elif ingress_relations:
+            relation = ingress_relations[0]
```
jdkandersson, line 88: Propose this interface:
```python
if nginx_route_relations:
    return IngressOption.from_nginx_route(nginx_route_relations[0])  # could also be via essence
if ingress_relations:
    return IngressOption.from_ingress_relations(ingress_relations[0], ingress_provider)  # could also be via essence
raise InternalError(...) # should not reach this point
```

src/charm.py
```diff
@@ -481,134 +43,61 @@ def __init__(self, *args) -> None:  # type: ignore[no-untyped-def]
...
-                    relation.app,
-                )
-                continue
-            dedup_ingress_relations.append(relation)
-        return tuple(nginx_route_relations + dedup_ingress_relations)
-
-    @property
-    def _all_config_or_relations(self) -> Any:
-        """Get all configuration and relation data.
+    def _get_ingress_options(self) -> Optional[IngressOption]:
```
jdkandersson, line 69: This should not return None, it should just return `IngressOptions`

jdkandersson: Perhaps separate the logic for whether the relation is present/ whether cleanup is needed

src/charm.py
```diff
@@ -621,19 +110,29 @@ def _namespace(self) -> Any:
         # for all instances. It would be very unusual for a relation to specify
         # this (arguably we should remove this as a relation option), so if set
         # via config it will be the same for all relations.
-        return self._all_config_or_relations[0]._namespace
+        return (
```
jdkandersson, line 113: Discussed this will be removed

src/charm.py
```diff
@@ -621,19 +110,29 @@ def _namespace(self) -> Any:
...
+            else self._ingress_options.service_namespace
+        )
 
     def _describe_ingresses_action(self, event: Any) -> None:
         """Handle the 'describe-ingresses' action.
 
         Args:
             event: Juju event that fires this handler.
         """
+        try:
```
jdkandersson, line 125: Discussed that we will remove this

src/charm.py
```diff
@@ -668,11 +176,13 @@ def _report_service_ips(self) -> List[str]:
         """
         api = self._core_v1_api()
         services = api.list_namespaced_service(  # type: ignore[attr-defined]
-            namespace=self._namespace
+            namespace=self._namespace, label_selector=self._label_selector
```
jdkandersson, line 179: Potentially remove in the future, leave for now

src/charm.py
```diff
@@ -683,539 +193,160 @@ def _report_ingress_ips(self) -> List[str]:
...
-            conf_or_rel: Ingress configuration or relation class.
+        Returns:
+            The name of the created or modified resource, None if no resource is
+            modified or created.
         """
-        api = self._core_v1_api()
-        body = conf_or_rel._get_k8s_service(self.app.name)
-        services = api.list_namespaced_service(  # type: ignore[attr-defined]
-            namespace=self._namespace
+        if self._ingress_options is None:
```
jdkandersson, line 228: `ingress_options` should be passed as an argument that shouldn't be `None`

src/charm.py
```diff
@@ -683,539 +193,160 @@ def _report_ingress_ips(self) -> List[str]:
...
+            return None
+        resource_list = controller.list_resource(
+            namespace=self._namespace, label_selector=self._label_selector
         )
-        if conf_or_rel._k8s_service_name in [x.metadata.name for x in services.items]:
-            api.patch_namespaced_service(  # type: ignore[attr-defined]
-                name=conf_or_rel._k8s_service_name,
+        body = controller.gen_resource_from_options(self._ingress_options)
+        if body.metadata.name in [r.metadata.name for r in resource_list]:
```
jdkandersson, line 234: this shouldn't use a list it should use a set

src/charm.py
```diff
@@ -481,159 +44,46 @@ def __init__(self, *args) -> None:  # type: ignore[no-untyped-def]
...
+            option_essence = IngressOptionEssence.from_nginx_route(
+                self, self.model.relations["nginx-route"][0]
+            )
+        elif self.model.relations["ingress"]:
+            relation = self.model.relations["ingress"][0]
+            if relation.app is None or not relation.data[relation.app]:
+                return None
+            option_essence = IngressOptionEssence.from_ingress(
+                self, self.model.relations["ingress"][0], self._ingress_provider
+            )
```
yanksyoon, line 80: I was wondering if this would be suitable for `state.py` (I've been using relations as state, I might be wrong though!).

src/charm.py
```diff
@@ -642,580 +92,247 @@ def k8s_auth(self) -> None:
...
+        ips = []
         for _ in range(count):
-            ingresses = api.list_namespaced_ingress(  # type: ignore[attr-defined]
-                namespace=self._namespace
+            ingresses = controller.list_resource(
+                namespace=ingress.metadata.namespace, label_selector=self._label_selector
             )
             try:
-                ips = [x.status.load_balancer.ingress[0].ip for x in ingresses.items]
+                ips = [x.status.load_balancer.ingress[0].ip for x in ingresses]
```
yanksyoon, line 137: From what I understand, it might be preferred to return a tuple rather than lists to keep it immutable.

yanksyoon: I was wondering if ips variable is needed, as in the ips can be returned right away if there are no `TypeError`s raised.

src/charm.py
```diff
@@ -642,580 +92,247 @@ def k8s_auth(self) -> None:
...
-            api.create_namespaced_ingress(
-                namespace=self._namespace,
-                body=body,
-            )
-            LOGGER.info(
-                "Ingress created in namespace %s with name %s",
-                self._namespace,
-                ingress_name,
+        nginx_route_relations = self.model.relations["nginx-route"]
+        ingress_relations = self.model.relations["ingress"]
```
yanksyoon, line 232: I think for relations that have `limit: 1`, it is better to use `self.model.get_relation(RELATION_NAME)`, since that targets a single relation.

src/charm.py
```diff
@@ -642,580 +92,247 @@ def k8s_auth(self) -> None:
...
-                self.unit.status = BlockedStatus(INVALID_BACKEND_PROTOCOL_MSG)
-                return
-        self.unit.status = ActiveStatus()
+    def _on_nginx_route_available(self, _: Any) -> None:
+        """Handle the nginx-route-available event."""
+        self._update_ingress()
+
+    def _on_nginx_route_broken(self, _: Any) -> None:
+        """Handle the nginx-route-broken event."""
+        self._update_ingress()
```
yanksyoon, line 335: I was wondering if in such cases where all events lead to reconcile, we can implement the [reconcile pattern](https://discourse.charmhub.io/t/eliminating-duplication-across-charms-by-using-a-reusable-reconcile-function/11304), I don't think it is currently approved internally within the team but just leaving it here quietly... 😂

src/controller.py
```diff
@@ -0,0 +1,585 @@
...
+AnyResource = typing.TypeVar(  # pylint: disable=invalid-name
+    "AnyResource",
+    kubernetes.client.V1Endpoints,
+    kubernetes.client.V1EndpointSlice,
+    kubernetes.client.V1Service,
+    kubernetes.client.V1Ingress,
+)
+
+
+class ResourceController(abc.ABC, typing.Generic[AnyResource]):
```
yanksyoon, line 26: I was wondering if a `Protocol` might be more suited here than ABCs, since from what I understand, we are trying to define interfaces for the charm to use.

yanksyoon (commented):
Learnt a lot of cool stuffs here, thanks!

src/charm.py
weiiwang01: There will only be one or two resource in the list, the performance gain is not very compelling in this context.

src/charm.py
weiiwang01: That's a good idea, updated, thanks!

src/charm.py
weiiwang01: Yes, I have received feedback suggesting a dedicated function handler with different names for each event...

src/controller.py
weiiwang01: Yes, the ABC and Protocol with explicitly declaring implementation are basically the same. It's more of a personal preference. Since we don't need implicit subtyping with structural typing, I just chose the ABC.

src/charm.py
weiiwang01: Since it's just one function for now, I guess we can put it here. If we are expanding this function in the future, we can move this to a different place. Thanks!

src/charm.py
weiiwang01: Updated, thanks!

src/charm.py
weiiwang01: Moved, thanks!

src/charm.py
```diff
@@ -481,134 +43,61 @@ def __init__(self, *args) -> None:  # type: ignore[no-untyped-def]
...
+                "remove follower units using `juju scale-application {self.app.name} 1`"
+            )
+            return
+
+        self._ingress_options: Optional[IngressOption] = None
+
+        self.framework.observe(self.on.config_changed, self._on_config_changed)
         self.framework.observe(self.on.describe_ingresses_action, self._describe_ingresses_action)
-        self.framework.observe(self.on.start, self._on_start)
+        self.framework.observe(self.on.start, self._on_config_changed)
```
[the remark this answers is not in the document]
weiiwang01: Updated, thanks!

src/charm.py
weiiwang01: Updated, thanks!

src/charm.py
weiiwang01: Updated, thanks!

src/charm.py
weiiwang01: Removed, thanks!

src/charm.py
weiiwang01: Removed, thanks!

src/charm.py
weiiwang01: Removed, thanks!

src/controller.py
weiiwang01: Actually, now I think the generic syntax of Protocol is more elegant, change to Protocol with explicitly declaring implementation. Thanks for the idea!

src/charm.py
```diff
@@ -642,580 +90,250 @@ def k8s_auth(self) -> None:
...
+        if options is not None:
+            service = define_resource(controller=service_controller, options=options)
+            ingress = define_resource(controller=ingress_controller, options=options)
+        cleanup_resources(controller=endpoints_controller, exclude=endpoints)
+        cleanup_resources(controller=endpoint_slice_controller, exclude=endpoint_slice)
+        cleanup_resources(controller=service_controller, exclude=service)
+        cleanup_resources(controller=ingress_controller, exclude=ingress)
+        return ingress
+
+    def _update_ingress(self) -> None:
```
jdkandersson, line 275: 1. Check preconditions -> error out if there is a problem
2. If there is no ingress relation, clean up everything and return
3. if there is a relation, take the data and convert it to IngressOptions
4. Create the ingress resources based on IngressOptions
5. cleanup any unnecessary ingress (e.g., created by the previous relation) that weren't just created

yanksyoon (dismissed):
Looks extremely configurable, LGTM!

src/charm.py
```diff
@@ -642,580 +107,251 @@ def k8s_auth(self) -> None:
...
-            except ConflictingRoutesError:
+            definition = self._get_definition_from_relation(relation)
+            ingress = self._reconcile(definition)
+            self.unit.status = WaitingStatus("Waiting for ingress IP availability")
+            ingress_ips = self._report_ingress_ips(ingress)
+            message = f"Ingress IP(s): {', '.join(ingress_ips)}" if ingress_ips else ""
+            self.unit.status = ActiveStatus(message)
+        except InvalidIngressError as exc:
+            self.unit.status = BlockedStatus(exc.msg)
+        except kubernetes.client.exceptions.ApiException as exception:
```
jdkandersson, line 319: It would be better to handle a more specific exception here, the k8s client is not used in this function and it raises a k8s exception which is confusing

src/charm.py
```diff
@@ -642,580 +107,251 @@ def k8s_auth(self) -> None:
...
+            if exception.status == 403:
+                LOGGER.error(
```
jdkandersson, line 321: Please update this excepion

src/charm.py
```diff
@@ -481,159 +45,60 @@ def __init__(self, *args) -> None:  # type: ignore[no-untyped-def]
...
+                model=self.model,
+                config=self.config,
+                relation=relation,
+                ingress_provider=self._ingress_provider,
+            )
+        ingress_definition = IngressDefinition.from_essence(definition_essence)
+        return ingress_definition
 
-    def k8s_auth(self) -> None:
+    def _k8s_auth(self) -> None:
```
jdkandersson, line 101: Let's move this to the constructor

src/charm.py
weiiwang01: Okay, I added a decorator in `controller.py` to remap exceptions.

src/charm.py
weiiwang01: Updated, thanks!

src/charm.py
```diff
@@ -481,159 +45,60 @@ def __init__(self, *args) -> None:  # type: ignore[no-untyped-def]
...
-        # We're querying the first one here because this will always be the same
-        # for all instances. It would be very unusual for a relation to specify
-        # this (arguably we should remove this as a relation option), so if set
-        # via config it will be the same for all relations.
-        return self._all_config_or_relations[0]._namespace
+            The current effective relation object, None if there are no relation.
+        """
+        if self.model.get_relation("nginx-route") is not None:
+            relation = cast(Relation, self.model.get_relation("nginx-route"))
+            if not (relation.app is None or not relation.data[relation.app]):
```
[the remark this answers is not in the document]
weiiwang01: Updated, thanks!

src/charm.py
weiiwang01: `kubernetes.config.load_incluster_config` has been moved to the charm constructor.

src/charm.py
```diff
@@ -642,580 +107,251 @@ def k8s_auth(self) -> None:
...
-        Returns:
-            The core v1 API.
-        """
-        self.k8s_auth()
-        return kubernetes.client.CoreV1Api()
-
-    def _networking_v1_api(self) -> kubernetes.client.NetworkingV1Api:
-        """Use the v1 beta1 networking API.
+    @property
+    def _endpoints_controller(self) -> EndpointsController:
```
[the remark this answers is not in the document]
weiiwang01: In that case, we need to systematically pass the `namespace` to the controller factory. I don't think that's a significant improvement. Also, I changed all the `*_controller` to be a simple instance attributes instead of property method, since we are moving the `kubernetes.config.load_incluster_config` to the charm class constructor. In that layout, passing the namespace to controller constructor will be difficult.

src/charm.py
```diff
@@ -642,580 +107,251 @@ def k8s_auth(self) -> None:
...
-        field_names = [f'_{f.replace("-", "_")}' for f in REQUIRED_INGRESS_RELATION_FIELDS]
-        return all(getattr(conf_or_rel, f) for f in field_names)
+    @property
+    def _label_selector(self) -> str:
+        """Get the label selector to select resources created by this app."""
+        return f"{CREATED_BY_LABEL}={self.app.name}"
 
-    def _delete_unused_services(self, current_svc_names: List[str]) -> None:
-        """Delete services and ingresses that are no longer used.
+    def _define_resource(
```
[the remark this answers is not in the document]
weiiwang01: The creation of `ResourceController` stems from the necessity for polymorphism across different resource types, not to introduce a new class or to shift functions from the charm class (that's also the reason why I try to maintain the controller's interface as closely aligned with the Kubernetes Python client as possible).

Personally, if the changes don't offer any significant improvement over the existing design, I don't want the `ResourceController` to assume the additional responsibilities.

src/charm.py
```diff
@@ -642,580 +107,251 @@ def k8s_auth(self) -> None:
...
+                "%s created in namespace %s with name %s",
+                controller.name,
+                namespace,
+                body.metadata.name,
             )
+        return body
 
-    def _remove_service(self, conf_or_rel: _ConfigOrRelation) -> None:
-        """Remove the created service in kubernetes.
+    def _cleanup_resources(
```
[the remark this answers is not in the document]
weiiwang01: See https://github.com/canonical/nginx-ingress-integrator-operator/pull/91#discussion_r1316827569, thanks!

src/charm.py
```diff
@@ -642,580 +107,251 @@ def k8s_auth(self) -> None:
...
+        endpoints_controller = self._endpoints_controller
+        endpoint_slice_controller = self._endpoint_slice_controller
+        service_controller = self._service_controller
+        ingress_controller = self._ingress_controller
+        namespace = definition.service_namespace if definition is not None else self.model.name
+        endpoints = None
+        endpoint_slice = None
+        define_resource = functools.partial(
+            self._define_resource, namespace=namespace, definition=definition
         )
```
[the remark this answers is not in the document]
weiiwang01: See https://github.com/canonical/nginx-ingress-integrator-operator/pull/91#discussion_r1316827569, thanks!

src/charm.py
```diff
@@ -642,580 +107,251 @@ def k8s_auth(self) -> None:
...
-        ]
-        all_additional_hostnames = [
-            conf_or_rel._additional_hostnames for conf_or_rel in self._all_config_or_relations
-        ]
-        for additional_hostname in all_additional_hostnames:
-            svc_hostnames.extend(additional_hostname)
-        self._delete_unused_ingresses(svc_hostnames)
-
-    def _on_config_changed(self, _: HookEvent) -> None:
+        cleanup_resources = functools.partial(self._cleanup_resources, namespace=namespace)
```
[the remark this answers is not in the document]
weiiwang01: See https://github.com/canonical/nginx-ingress-integrator-operator/pull/91#discussion_r1316827569, thanks!

src/charm.py
```diff
@@ -642,580 +107,251 @@ def k8s_auth(self) -> None:
...
+        ingress = define_resource(controller=ingress_controller)
+        cleanup_resources(controller=endpoints_controller, exclude=endpoints)
+        cleanup_resources(controller=endpoint_slice_controller, exclude=endpoint_slice)
+        cleanup_resources(controller=service_controller, exclude=service)
+        cleanup_resources(controller=ingress_controller, exclude=ingress)
+        return ingress
+
+    def _cleanup(self) -> None:
+        """Cleanup all resources managed by the charm."""
+        cleanup_resources = functools.partial(self._cleanup_resources, namespace=self.model.name)
```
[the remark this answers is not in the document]
weiiwang01: See https://github.com/canonical/nginx-ingress-integrator-operator/pull/91#discussion_r1316827569, thanks!

src/charm.py
```diff
@@ -481,741 +46,276 @@ def __init__(self, *args) -> None:  # type: ignore[no-untyped-def]
...
+        cleanup_resources(controller=self._endpoints_controller)
+        cleanup_resources(controller=self._endpoint_slice_controller)
```
jdkandersson, line 273: EndpointController(...).cleanup()
EndpointController(...).create()

src/charm.py
```diff
@@ -481,741 +43,243 @@ def __init__(self, *args) -> None:  # type: ignore[no-untyped-def]
...
-        )
-        all_k8s_service_names = [rel._k8s_service_name for rel in self._all_config_or_relations]
-        return [
-            x.spec.cluster_ip for x in services.items if x.metadata.name in all_k8s_service_names
-        ]
+        if relation.name == "nginx-route":
+            definition_essence = IngressDefinitionEssence(
+                model=self.model, config=self.config, relation=relation
+            )
+        else:
```
jdkandersson, line 135: Please change this to check that the relation name is `ingress` and have an else condition that raises an error for defensive programming purposes

src/charm.py
```diff
@@ -481,741 +43,243 @@ def __init__(self, *args) -> None:  # type: ignore[no-untyped-def]
...
+        nginx_route_relation = self.model.get_relation("nginx-route")
+        ingress_relation = self.model.get_relation("ingress")
+        if nginx_route_relation is not None and ingress_relation is not None:
+            raise InvalidIngressError(
+                "nginx-ingress-integrator cannot establish more than one relation at a time"
             )
 
-    def _remove_ingress(self, ingress_name: str) -> None:
-        """Remove ingress resource.
+    def _reconcile(self, definition: IngressDefinition) -> kubernetes.client.V1Ingress:
```
jdkandersson, line 201: This shouldn't return a k8s client object, let's find a way to not need this return value, e.g., retrieving it later

src/charm.py
```diff
@@ -481,741 +43,243 @@ def __init__(self, *args) -> None:  # type: ignore[no-untyped-def]
...
+        controller = self._get_ingress_controller(namespace=ingress.metadata.namespace)
         # Wait up to `interval * count` seconds for ingress IPs.
-        count, interval = _report_interval_count(), 1
+        count, interval = 100, 1
+        ips = []
         for _ in range(count):
-            ingresses = api.list_namespaced_ingress(  # type: ignore[attr-defined]
-                namespace=self._namespace
-            )
+            ingresses = controller.list_resource()
```
jdkandersson, line 159: let's change this to have a method to get the IP addresses rather than expose the underlying data structure

weiiwang01: Updated, thanks!

src/charm.py
weiiwang01: Updated, thanks!

src/charm.py
weiiwang01: Exhaustiveness checking added, thanks!

pyproject.toml
```diff
@@ -11,7 +11,7 @@ skips = ["*/*test.py", "*/test_*.py"]
 branch = true
 
 [tool.coverage.report]
-fail_under = 92
+fail_under = 90
```
jdkandersson, line 14: Hmm, this is going backwards which we have a standard which doesn't allow for that

pyproject.toml
```diff
@@ -43,7 +43,11 @@ select = ["E", "W", "F", "C", "N", "R", "D", "H"]
 # Ignore D107 Missing docstring in __init__
 ignore = ["W503", "E501", "D107"]
 # D100, D101, D102, D103: Ignore missing docstrings in tests
-per-file-ignores = ["tests/*:D100,D101,D102,D103,D104,D205,D212,D415,DC0020"]
+per-file-ignores = [
+    "src/ingress_definition.py:DCO060",
+    "src/controller.py:DCO020,DCO030,DCO031,DCO050,DCO055,DCO060",
```
jdkandersson, line 48: Could you check whether these are still needed?

src/controller.py
```diff
@@ -0,0 +1,790 @@
...
+import typing
+
+import kubernetes.client
+
+from exceptions import InvalidIngressError
+from ingress_definition import IngressDefinition
+
+logger = logging.getLogger(__name__)
+
+AnyResource = typing.TypeVar(  # pylint: disable=invalid-name
```
jdkandersson, line 19: What is pylint expecting here please? Any linting disable should include a comment with the reason why

pyproject.toml
weiiwang01: Gotcha, I have added some more tests and set the threshold to 92 again.

pyproject.toml
weiiwang01: Okay, I have checked, and the `DCO020` `DCO030` and `DCO060` can be removed. But the rest has something to do with the abstract method don't have a `return` statement, and the problem with re-raising exception with a `raise` statement.

src/controller.py
weiiwang01: Actually this is not required, don't know why I added this here, removed, thanks!

jdkandersson (approved):

claudiubelu:
From what I can tell, this basically introduces regressions when it comes to Nginx Ingress.

This limits the ingress relation to 1, which is opposite to what was requested by a user [1]. Secondly, it also bricks the usage of ``nginx/nginx-ingress`` Controller typically used in EKS deployments [2]:

```
The nginx/nginx-ingress Controller typically used in EKS doesn't handle
multiple Kubernetes Ingress Resources having the same hostname in the same
way as the k8s.gcr.io/ingress-nginx/controller:v1.1.0 Controller used in
microk8s. The former doesn't handle them at all, unless you specify certain
annotations, and an additional Kubernetes Ingress Resource [1]. But both
Controllers can handle one Kubernetes Ingress Resource containing multiple
hosts / routes.
```

Basically, multiple relations were introduced so **only one** Ingress resource would be created, which would contain all the necessary routes associated with a **single** service hostname, allowing us to bundle different services under the **same** name.

Additionally, limiting to only one relation per charm, would mean adding N instances of a charm for N charms that would require that relation, instead of having just one ``nginx-ingress-integrator`` charm. That makes it more of a hassle than it should to maintain and update. Plus, IMO, it makes deployments look messier and unappealing for future clients and potential adopters of Juju.

Hopefully these issues will be addressed somehow before moving forward.

[1] https://bugs.launchpad.net/charm-k8s-ingress/+bug/1955597
[2] https://git.launchpad.net/charm-k8s-ingress/commit/?id=59f1ca1b4dec1a3fc7cb659b554800ae15b943d3

weiiwang01:
> From what I can tell, this basically introduces regressions when it comes to Nginx Ingress.
> 
> This limits the ingress relation to 1, which is opposite to what was requested by a user [1]. Secondly, it also bricks the usage of `nginx/nginx-ingress` Controller typically used in EKS deployments [2]:
> 
> ```
> The nginx/nginx-ingress Controller typically used in EKS doesn't handle
> multiple Kubernetes Ingress Resources having the same hostname in the same
> way as the k8s.gcr.io/ingress-nginx/controller:v1.1.0 Controller used in
> microk8s. The former doesn't handle them at all, unless you specify certain
> annotations, and an additional Kubernetes Ingress Resource [1]. But both
> Controllers can handle one Kubernetes Ingress Resource containing multiple
> hosts / routes.
> ```
> 
> Basically, multiple relations were introduced so **only one** Ingress resource would be created, which would contain all the necessary routes associated with a **single** service hostname, allowing us to bundle different services under the **same** name.
> 
> Additionally, limiting to only one relation per charm, would mean adding N instances of a charm for N charms that would require that relation, instead of having just one `nginx-ingress-integrator` charm. That makes it more of a hassle than it should to maintain and update. Plus, IMO, it makes deployments look messier and unappealing for future clients and potential adopters of Juju.
> 
> Hopefully these issues will be addressed somehow before moving forward.
> 
> [1] https://bugs.launchpad.net/charm-k8s-ingress/+bug/1955597 [2] https://git.launchpad.net/charm-k8s-ingress/commit/?id=59f1ca1b4dec1a3fc7cb659b554800ae15b943d3

@claudiubelu Thanks for your input! The rationale behind this change is that if multiple relations are connected to the same `nginx-ingress-integrator` charm, then the charm configuration will be applied to all the relations. This becomes especially tricky for configurations like `tls-secret-name`. For that reason, more often in our deployment, we simply deploy an `nginx-ingress-integrator` charm for each application. Since the `nginx-ingress-integrator` charm is a workload-less charm, there isn't too much overhead. And now, as we are some plans to add the `tls-certificates` relation to the `nginx-ingress-integrator` charm, it will further complicate the situation. We have also received user request to create multiple ingress resources for the same host to add different annotations to different paths: https://github.com/canonical/nginx-ingress-integrator-operator/issues/78. It's quite a challenge to satisfy both the Nginx's Ingress Controller users and Kubernetes' Nginx Ingress Controller users.

We are aware this is a significant breaking change. That's why this will not be merged into the main branch of the `nginx-ingress-integrator`. Instead, we will release the new version in a different channel on the charmhub, likely called `v2`.

weiiwang01:
All progress moved to the `v2` branch, close this pull request.

Issue #90, opened 2023-08-16
Uncaught exception while removing ingress relation


claudiubelu:
### Bug Description

We're currently using the ``nginx-ingress-integrator`` charm (revision 79) for multiple charms. However, when we removed the ``ingress`` relation between it and another charm, the charm ended up in an error state (``hook failed: "ingress-relation-broken"``), with the stack trace seen below in ``juju debug-log``.

### To Reproduce

1. Deploy 2 charms with an ``ingress`` ``requires`` interface.
2. Deploy ``nginx-ingress-integrator`` charm (edge, rev 79).
3. Relate the 2 charms to ``nginx-ingress-integrator``.
4. Remove a relation.

### Environment

Juju version: 3.1.5
Environment: EKS

### Relevant log output

```shell
unit-ingress-0: 10:17:09 ERROR unit.ingress/0.juju-log ingress:58: Uncaught exception while in charm code:
Traceback (most recent call last):
  File "./src/charm.py", line 1220, in <module>
    main(NginxIngressCharm)
  File "/var/lib/juju/agents/unit-ingress-0/charm/venv/ops/main.py", line 441, in main
    _emit_charm_event(charm, dispatcher.event_name)
  File "/var/lib/juju/agents/unit-ingress-0/charm/venv/ops/main.py", line 149, in _emit_charm_event
    event_to_emit.emit(*args, **kwargs)
  File "/var/lib/juju/agents/unit-ingress-0/charm/venv/ops/framework.py", line 344, in emit
    framework._emit(event)
  File "/var/lib/juju/agents/unit-ingress-0/charm/venv/ops/framework.py", line 841, in _emit
    self._reemit(event_path)
  File "/var/lib/juju/agents/unit-ingress-0/charm/venv/ops/framework.py", line 930, in _reemit
    custom_handler(event)
  File "/var/lib/juju/agents/unit-ingress-0/charm/lib/charms/nginx_ingress_integrator/v0/ingress.py", line 395, in _on_relation_broken
    self.charm.on.ingress_broken.emit(event.relation)
  File "/var/lib/juju/agents/unit-ingress-0/charm/venv/ops/framework.py", line 344, in emit
    framework._emit(event)
  File "/var/lib/juju/agents/unit-ingress-0/charm/venv/ops/framework.py", line 841, in _emit
    self._reemit(event_path)
[...]

PR #92, opened 2023-08-18
Ignore broken relations during relation-broken events

weiiwang01:
When the `relation-broken` event is triggered, there's a possibility that the broken relation, although listed, the remote application may not be accessible. This update ensures such relations are safely ignored.

Fix: https://github.com/canonical/nginx-ingress-integrator-operator/issues/90


arturo-seijas (approved):
LGTM

jdkandersson (approved):

PR #94, opened 2023-09-01
Add nginx-route how-to docs

gtrkiller:
Add nginx-route relation how-to to the charm.

### DIFF
--- a/docs/how-to/add-the-ingress-relation.md
+++ b/docs/how-to/add-the-ingress-relation.md
@@ -1,6 +1,6 @@
 # How to add the Ingress relation.
 
-The ingress relation is preferred over the nginx-route relation if you want to use generic features. If you need
+The `ingress` relation is preferred over the `nginx-route` relation if you want to use generic features. If you need
 something nginx-specific such as owasp-modsecurity-crs, then please follow the [nginx-route relation](https://charmhub.io/nginx-ingress-integrator/docs/add-the-nginx-route-relation) tutorial instead.
 
 ## Requirements
@@ -40,7 +40,7 @@ Now let's deploy the charm just to confirm everything is working as expected wit
 ```
 charmcraft pack
 ```
-This will build the charm inside an LXC for you. The output will tell you the location of the built charm. In my case, that was `my-charm_ubuntu-22.04-amd64.charm`.
+This will build the charm inside an LXC container for you. The output will tell you the location of the built charm. For example, `my-charm_ubuntu-22.04-amd64.charm`.
 
 Now let's add a juju model and deploy our charm.
 ```
@@ -63,7 +63,7 @@ my-charm/0*  active    idle   10.1.129.139
 
 So we now have a working charm, great! However, what we don't currently have is ingress for our application configured. MicroK8s sets up networking in such a way that you can reach the IPs of units directly, but in a production Kubernetes cluster things don't work in this way. Also, we're visiting a cluster-internal IP address directly, what if we want a real hostname/IP address for this? We'll need to configure Ingress for this to be possible.
 
-Also, you may notice that you can visit the Unit IP address in a browser, but not the App IP address. Why is that? The reason is that the App IP address refers to a [Kubernetes Service](https://kubernetes.io/docs/concepts/services-networking/service/) so as you add other units to the charm they would in theory be reachable through the same IP. However, Juju doesn't yet have a mechanism for a charm to define what port that Service should be configured with, and this why we can't use it to browse the web site (which is listening on port 80).
+Also, you may notice that you can visit the Unit IP address in a browser, but not the App IP address. Why is that? The reason is that the App IP address refers to a [Kubernetes Service](https://kubernetes.io/docs/concepts/services-networking/service/) so as you add other units to the charm they would in theory be reachable through the same IP. However, Juju doesn't have a mechanism for a charm to define what port that Service should be configured with, and this why we can't use it to browse the web site (which is listening on port 80).
 
 ## Add the Ingress relation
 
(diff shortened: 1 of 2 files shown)


docs/how-to/add-the-ingress-relation.md
```diff
@@ -1,5 +1,8 @@
 # How to add the Ingress relation.
 
+The ingress relation is preferred over the nginx-route relation if you want to use generic features. If you need
```
jdkandersson, line 3: do you think it is worth making `nginx-route` and `ingress` quoted with `

docs/how-to/add-the-nginx-route-relation.md
```diff
@@ -0,0 +1,146 @@
...
+
+We can leave the other files as is for now, but if we want to publish this charm later, we'll want to update them before doing so.
+
+## Deploy the charm
+
+Now let's deploy the charm just to confirm everything is working as expected without the relation. To do that, just run this:
+```
+charmcraft pack
+```
+This will build the charm inside an LXC for you. The output will tell you the location of the built charm. In my case, that was `my-charm_ubuntu-22.04-amd64.charm`.
```
jdkandersson, line 43: perhaps "LXC container"?

jdkandersson: Perhaps avoid referring to "my case" and instead just say, for example

docs/how-to/add-the-nginx-route-relation.md
```diff
@@ -0,0 +1,146 @@
...
+
+Unit         Workload  Agent  Address       Ports  Message
+my-charm/0*  active    idle   10.1.129.139         
+```
+
+So we now have a working charm, great! However, what we don't currently have is ingress for our application configured. MicroK8s sets up networking in such a way that you can reach the IPs of units directly, but in a production Kubernetes cluster things don't work in this way. Also, we're visiting a cluster-internal IP address directly, what if we want a real hostname/IP address for this? We'll need to configure Ingress for this to be possible.
+
+Also, you may notice that you can visit the Unit IP address in a browser, but not the App IP address. Why is that? The reason is that the App IP address refers to a [Kubernetes Service](https://kubernetes.io/docs/concepts/services-networking/service/) so as you add other units to the charm they would in theory be reachable through the same IP. However, Juju doesn't yet have a mechanism for a charm to define what port that Service should be configured with, and this why we can't use it to browse the web site (which is listening on port 80).
```
jdkandersson, line 66: by saying juju doesn't have this feature yet, are we implying that it is coming in the future?

gregory-schiano:
This PR will have to wait for https://github.com/canonical/nginx-ingress-integrator-operator/pull/91 to be merged before being merged as it mentions `nginx-ingress-integrator` for the ingress relation

docs/how-to/add-the-nginx-route-relation.md
gregory-schiano: Since that's a copy paste of the now renamed `ingress` relation documentation, any changes made here should be reflected on the other documentation

arturo-seijas (dismissed):

docs/how-to/add-the-ingress-relation.md
```diff
@@ -63,15 +63,15 @@ my-charm/0*  active    idle   10.1.129.139
...
+Also, you may notice that you can visit the Unit IP address in a browser, but not the App IP address. Why is that? The reason is that the App IP address refers to a [Kubernetes Service](https://kubernetes.io/docs/concepts/services-networking/service/) so as you add other units to the charm they would in theory be reachable through the same IP. However, Juju doesn't have a mechanism for a charm to define what port that Service should be configured with, and this why we can't use it to browse the web site (which is listening on port 80).
 
 ## Add the Ingress relation
 
 First of all, let's grab the [relation library](https://charmhub.io/nginx-ingress-integrator/libraries/ingress). We can do this by running this:
 ```
 charmcraft fetch-lib charms.nginx_ingress_integrator.v0.ingress
 ```
-This has downloaded `lib/charms/nginx_ingress_integrator/v0/ingress.py`. Now we just need to update `src/charm.py`.
+This has downloaded `lib/charms/nginx_ingress_integrator/v0/ingress.py`. Now we just need to update `/charm.py`.
```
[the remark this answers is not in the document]
gtrkiller: I might have deleted that accidentally, thank you for spotting that!

docs/how-to/add-the-nginx-route-relation.md
```diff
@@ -0,0 +1,146 @@
...
+As you can see, we're adding support for a configuration option of `external-hostname` that will be used when configuring `nginx-route`. Let's update `config.yaml` to enable this. Add the following the end of that file:
+```
+  external-hostname:
+    description: |
+      The external hostname to use. Will default to the name of the deployed
+      application.
+    default: ""
+    type: string
+```
+Now we just need to add the relation definition to `metadata.yaml`. Add the following to the end of that file:
```
amandahla, line 99: nit-picky: It seems there are too many "Now"s

jdkandersson (approved):

docs/how-to/add-the-ingress-relation.md
```diff
@@ -60,15 +63,15 @@ my-charm/0*  active    idle   10.1.129.139
...
 
-Also, you may notice that you can visit the Unit IP address in a browser, but not the App IP address. Why is that? The reason is that the App IP address refers to a [Kubernetes Service](https://kubernetes.io/docs/concepts/services-networking/service/) so as you add other units to the charm they would in theory be reachable through the same IP. However, Juju doesn't yet have a mechanism for a charm to define what port that Service should be configured with, and this why we can't use it to browse the web site (which is listening on port 80).
+Also, you may notice that you can visit the Unit IP address in a browser, but not the App IP address. Why is that? The reason is that the App IP address refers to a [Kubernetes Service](https://kubernetes.io/docs/concepts/services-networking/service/) so as you add other units to the charm they would in theory be reachable through the same IP. However, Juju doesn't have a mechanism for a charm to define what port that Service should be configured with, and this why we can't use it to browse the web site (which is listening on port 80).
```
weiiwang01, line 66: Kubernetes applications should now have the ability to define ports in the service using `open-port`. However, it appears that this feature is exclusive to Juju 3. https://github.com/juju/juju/pull/14975

weiiwang01 (approved):

Issue #93, opened 2023-08-24
ExternalName ingress route support


claudiubelu:
### Enhancement Proposal

Kubernetes has a Service type ``ExternalName`` (https://kubernetes.io/docs/concepts/services-networking/service/#externalname), which allows the Cluster DNS Service to return a CNAME record to the set ``externalName``. This can be useful in certain situations, such as creating a ingress routes for a github page (https://pages.github.com/).

The ``nginx-ingress-integrator`` does not have any support for this at this moment. The charm has no config options for this. As for charms related to it, the charm creates ``ClusterIP`` services and references them in the ``ingress`` objects it creates. Because the charm itself creates the ``ClusterIP`` service used for Ingress, we can't simply develop another workload-less charm that would define the ``ExternalName`` Service and relate it to the ``nginx-ingress-integrator`` charm and have it use it in the ingress definition.

For addressing this, there are few solutions:

- Add an ``external-name`` config option and relation data field. Setting this would imply that the Service type will be ``ExternalName``. The charm would then create the Kubernetes service with that type, and set the appropiate ``externalName`` field in its spec. The downside for this: Let's say that we'd want ``foo.lish.com/`` to point towards some github pages, however we already have an ``nginx-ingress-integrator`` charm related to our application charm (for ``foo.lish.com/bar``). This means that we can't configure the same charm to create the ``ExternalName``, as the config options override the relation data and do not represent a separate entity. We can't simply deploy a second ``nginx-ingress-integrator`` charm and configure that with the ``externalName`` and the same ``service-hostname`` as we can't have 2 ``nginx-ingress-integrator`` charms manage the same Ingress resource, and we can't have 2 separate Ingress resources with the same hosts (not all NGINX Ingress Controllers supports this). A solution for this would be to create a skeleton charm, which basically only has the ingress relation and configures it.

- Add CRUD actions for ingress routes.

weiiwang01:
Another problem is that I think the external name requires the `nginx.ingress.kubernetes.io/backend-protocol` set to `HTTPS` on the ingress resources to forward HTTPS traffic. This setting could conflict with other services connected to the ingress.

[...]

PR #96, opened 2023-09-21
Adds ExternalName Service type option

claudiubelu:
Kubernetes has a Service type `ExternalName`, which allows the Cluster DNS Service to return a CNAME record to the set  `externalName`. This can be useful in certain situations.

The `nginx-ingress-integrator` charm does not have any support for this at this moment.

This commit adds the `external-name` relation field and config option, which, if set, will cause the `nginx-ingress-integrator` charm to create a Kubernetes ExternalName Service with the `externalName` set instead of a ClusterIP Service.

Implements: https://github.com/canonical/nginx-ingress-integrator-operator/issues/93

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -94,6 +94,12 @@ options:
     default: ""
     description: The path to rewrite requests to. If not set, rewrite-target will be "/".
     type: string
+  external-name:
+    default: ""
+    description: |
+      The external name of the service to create an ingress for.
+      The charm will create a Kubernetes ExternalName Service instead of a ClusterIP, allowing kubedns or equivalent to return it as a CNAME record.
+    type: string
   service-hostname:
     default: ""
     description: The hostname of the service to create an ingress for.
--- a/lib/charms/nginx_ingress_integrator/v0/ingress.py
+++ b/lib/charms/nginx_ingress_integrator/v0/ingress.py
@@ -12,6 +12,7 @@ Import `IngressRequires` in your charm, with two required options:
 `config_dict` accepts the following keys:
 - additional-hostnames
 - backend-protocol
+- external-name
 - limit-rps
 - limit-whitelist
 - max-body-size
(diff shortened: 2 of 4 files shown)


cbartz:
/canonical/self-hosted-runners/run-workflows d560067e7bf15dae02b4d5638eba3900990e51a4

arturo-seijas (approved):
LGTM, thx!

jdkandersson:
@weiiwang01 could you please take a look at this

weiiwang01:
> @weiiwang01 could you please take a look at this

@claudiubelu  Sorry for the delay, I've been doing some experiments related to this. I have some concerns regarding the use of an external name with a HTTPS backend, as it seems to be a major use case. By default, Nginx ingress does not validate the certificate of the HTTPS backend, which could pose significant security risks. And make it worse, configuring Nginx ingress to work with a HTTPS backend and ensuring that it validates the backend is not straightforward at all.

Here's an example Kubernetes manifest file using an external name to connect to a website with a self-signed certificate. And the Nginx ingress doesn't complain at all.

```yaml
# curl -H 'Host: self-signed.badssl.com' http://127.0.0.1 # no complain
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: example-ingress
  annotations:
    nginx.ingress.kubernetes.io/backend-protocol: HTTPS
spec:
  ingressClassName: public
  rules:
  - host: "self-signed.badssl.com"
    http:
      paths:
      - pathType: Prefix
        path: "/"
        backend:
          service:
            name: external-name-example
            port:
              number: 443
---
apiVersion: v1
kind: Service
metadata:
  name: external-name-example
  labels:
    endpoint-group: example
spec:
  type: ExternalName
  externalName: self-signed.badssl.com
```

arturo-seijas (commented):
/canonical/self-hosted-runners/run-workflows 04b5d490c8c93b26d1c72afc987be10a974e4098

cbartz:
/canonical/self-hosted-runners/run-workflows 04b5d490c8c93b26d1c72afc987be10a974e4098

weiiwang01:
@claudiubelu I think for now we should close this pull request considering the current security concerns. I think this is more of an issue of the Kubernetes Nginx ingress controller. But we should definitely revisit this after we have a solution for enabling the HTTP security.

PR #101, opened 2023-10-31
Standard Ingress tls relation

gtrkiller:
Implement TLS relation for v2 Ingress charm.


src/charm.py
```diff
@@ -252,6 +326,198 @@ def _on_nginx_route_broken(self, _: Any) -> None:
...
+        relation = self._get_relation()
+        if relation is None:
+            self._cleanup()
+            self.unit.status = WaitingStatus("waiting for relation")
+            return
+        definition = self._get_definition_from_relation(relation)
+        subject = definition.service_hostname if definition is not None else self.model.name
+        csr = generate_csr(
+            private_key=secret.get_content()["key"].encode(),
+            private_key_password=secret.get_content()["password"].encode(),
```
weiiwang01, line 399: If we are using juju secrets to store the private keys, it means that we might not be able to deploy the nginx-ingress-integrator in juju 2.9 (PS5). Maybe we should consider using the Kubernetes secrets, as the Kubernetes TLS secrets can contain the key with an empty cert and some extra fields like csr.

weiiwang01 (commented):
I've just some of the code and noticed something. In the `tls-certificates` library example, there are two distinct relations. One is a peer relation named `replicas`, which is used for storing information exclusive to the requirer, like the private key, the most recently used CSR, and so forth. The second relation, `certificates`, is the one for communication with the `tls-certificates` provider. However, in the current implementation, these two relations appear to be mixed with each other.

I think in the context of the `nginx-ingress-integrator` charm, the peer relation from the example might be the Kubernetes TLS secrets. The Kubernetes TLS secrets aren't just for certificates and keys; they can store additional data too. So, we can store items like the latest used CSR, CA certificates, and more within the Kubernetes TLS secret.

.licenserc.yaml
```diff
@@ -30,6 +30,7 @@ header:
     - 'CODEOWNERS'
     - 'icon.svg'
     - 'LICENSE'
+    - 'tests/integration/provider_charm/**'
```
arturo-seijas, line 33: Why do you need to exclude this?

gtrkiller: That folder's content is taken literally from https://github.com/canonical/tls-certificates-interface/tree/main/tests/unit/charms/tls_certificates_interface/v2/dummy_provider_charm, it is not of my making. So I'd like to conserve the code as similar to the source as possible.

arturo-seijas: Isn't there a charm implementing the relation you can just deploy? You are duplicating code across repositories. If not, you can also use AnyCharm

src/charm.py
```diff
@@ -336,12 +336,12 @@ def _on_get_certificate_action(self, event: ActionEvent) -> None:
         if not tls_certificates_relation:
             event.fail("Certificates relation not created.")
             return
-        if tls_certificates_relation.data[self.unit]["certificate"]:
+        if tls_certificates_relation.data[self.app]["certificate"]:
```
arturo-seijas, line 339: Nitpick, this will be easier to to read if you extract `tls_certificates_relation.data[self.app]` to a variable. Same for the rest of the methods

src/charm.py
```diff
@@ -252,6 +326,198 @@ def _on_nginx_route_broken(self, _: Any) -> None:
...
+    def _on_certificates_relation_created(self, event: RelationCreatedEvent) -> None:
+        """Handle the TLS Certificate relation created event.
+
+        Args:
+            event: The event that fires this method.
+        """
+        tls_certificates_relation = self.model.get_relation("certificates")
+        if not tls_certificates_relation:
+            self.unit.status = WaitingStatus("Waiting for peer relation to be created")
+            event.defer()
```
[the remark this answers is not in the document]
gtrkiller: I'd prefer to defer to avoid checking password existance multiple times unnecessarilly

src/charm.py
```diff
@@ -252,6 +333,312 @@ def _on_nginx_route_broken(self, _: Any) -> None:
...
+            return
+        relation = self._get_relation()
+        if relation is None:
+            self._cleanup()
+            self.unit.status = WaitingStatus("waiting for relation")
+            return
```
gregory-schiano, line 427: This precondition check is repeated multiple times, I think some factoring could help

src/charm.py
```diff
@@ -252,6 +333,312 @@ def _on_nginx_route_broken(self, _: Any) -> None:
...
+            )
+            self._tls.update_relation_data_fields(
+                {f"csr-{hostname}": csr.decode()}, tls_certificates_relation, self.app
+            )
+            peer_relation = self.model.get_relation("nginx-peers")
+            if peer_relation:
+                self._tls.update_relation_data_fields(
+                    {f"csr-{hostname}": csr.decode()}, peer_relation, self.app
+                )
+            self.certificates.request_certificate_creation(certificate_signing_request=csr)
```
gregory-schiano, line 456: Shouldn't all this business logic be in the TLS service ?

src/charm.py
```diff
@@ -252,6 +333,312 @@ def _on_nginx_route_broken(self, _: Any) -> None:
...
+                    secret.set_content(private_key_dict)
+                except SecretNotFoundError:
+                    secret = self.app.add_secret(
+                        content=private_key_dict, label=f"private-key-{hostname}"
+                    )
+            self._tls.update_relation_data_fields(
+                private_key_dict, tls_certificates_relation, self.app
+            )
+            peer_relation = self.model.get_relation("nginx-peers")
+            if peer_relation:
```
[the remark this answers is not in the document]
gregory-schiano: Question is valid also every time we do `if peer_relation:`

src/charm.py
```diff
@@ -252,6 +333,312 @@ def _on_nginx_route_broken(self, _: Any) -> None:
...
+        private_key = ""
+        if JujuVersion.from_environ().has_secrets:
+            secret = self.model.get_secret(label=f"private-key-{hostname}")
+            private_key = secret.get_content()["key"]
+        else:
+            private_key = self._tls.get_relation_data_field(
+                f"key-{hostname}", tls_certificates_relation, self.app
+            )
+        self._tls.cert[hostname] = event.certificate
+        self._tls.key[hostname] = private_key
```
gregory-schiano, line 502: Shouldn't all this business logic be in TLS service ?

src/charm.py
```diff
@@ -252,6 +333,312 @@ def _on_nginx_route_broken(self, _: Any) -> None:
...
+            new_certificate_signing_request=new_csr,
+        )
+        self._tls.update_relation_data_fields(
+            {f"csr-{hostname}": new_csr.decode()}, tls_certificates_relation, self.app
+        )
+        peer_relation = self.model.get_relation("nginx-peers")
+        if peer_relation:
+            self._tls.update_relation_data_fields(
+                {f"csr-{hostname}": new_csr.decode()}, peer_relation, self.app
+            )
```
gregory-schiano, line 560: Shouldn't all this business logic be in TLS service ?

src/charm.py
```diff
@@ -252,6 +333,312 @@ def _on_nginx_route_broken(self, _: Any) -> None:
...
+            self.unit.status = WaitingStatus("Waiting for peer relation to be created")
+            event.defer()
+            return
+        if event.reason == "revoked":
+            hostname = self._tls.get_hostname_from_csr(
+                tls_certificates_relation, self.app, event.certificate_signing_request
+            )
+            self._certificate_revoked([hostname])
+        if event.reason == "expired":
+            self._on_certificate_expiring(event)
```
gregory-schiano, line 625: I'd rather avoid calling a hoot method directly. Put the common business logic of both event in a private method and call it from both hook method, like you did with `_certificate_revoked`

src/charm.py
```diff
@@ -252,6 +333,312 @@ def _on_nginx_route_broken(self, _: Any) -> None:
...
+            self._cleanup()
+            self.unit.status = WaitingStatus("waiting for relation")
+            return
+        for hostname in revoke_list:
+            old_csr = self._tls.get_relation_data_field(
+                f"csr-{hostname}", tls_certificates_relation, self.app
+            )
+            if not old_csr:
+                continue
+            if JujuVersion.from_environ().has_secrets:
```
gregory-schiano, line 584: Shouldn't we managed the case when juju doesn't support secrets ?

src/charm.py
```diff
@@ -252,6 +333,312 @@ def _on_nginx_route_broken(self, _: Any) -> None:
...
+            return
+        hostnames = self.get_additional_hostnames()
+        for hostname in hostnames:
+            private_key_password = self._tls.generate_password().encode()
+            private_key = generate_private_key(password=private_key_password)
+            private_key_dict = {
+                "password": private_key_password.decode(),
+                "key": private_key.decode(),
+            }
+            if JujuVersion.from_environ().has_secrets:
```
gregory-schiano, line 397: Shouldn't we manage the case when juju doesn't support secrets ?

src/charm.py
```diff
@@ -252,6 +333,312 @@ def _on_nginx_route_broken(self, _: Any) -> None:
...
+        Args:
+            _: The event that fires this method.
+        """
+        hostnames = self.get_additional_hostnames()
+        for hostname in hostnames:
+            try:
+                secret = self.model.get_secret(label=f"private-key-{hostname}")
+                secret.remove_all_revisions()
+            except SecretNotFoundError:
+                LOGGER.warning("Juju secret for %s already does not exist", hostname)
```
gregory-schiano, line 640: Shouldn't we manage the case when juju doesn't support secrets ?

src/controller.py
```diff
@@ -1,6 +1,6 @@
 # Copyright 2023 Canonical Ltd.
 # See LICENSE file for licensing details.
-
+# pylint: disable=too-many-lines
```
gregory-schiano, line 3: This is probably a sign that we need to split in different module each controller

src/controller.py
```diff
@@ -189,7 +190,9 @@ def cleanup_resources(
             )
 
 
-class EndpointsController(ResourceController[kubernetes.client.V1Endpoints]):
+class EndpointsController(
+    ResourceController[kubernetes.client.V1Endpoints]
+):  # pylint: disable=inherit-non-class
```
gregory-schiano, line 195: Why do we need this now ?

gregory-schiano (commented):
Didn't review the tests nor `tls_relation.py` fully, as I feel some adaptation on the service interface is required

src/controller.py
gtrkiller: I agree, but I think it is out of scope for this PR, maybe opening another PR with refactoring will do, what do you think?

src/charm.py
```diff
@@ -252,6 +333,312 @@ def _on_nginx_route_broken(self, _: Any) -> None:
...
+        return hostnames
+
+    def _on_certificates_relation_created(self, event: RelationCreatedEvent) -> None:
+        """Handle the TLS Certificate relation created event.
+
+        Args:
+            event: The event that fires this method.
+        """
+        tls_certificates_relation = self.model.get_relation("certificates")
+        if not tls_certificates_relation:
```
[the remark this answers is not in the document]
gtrkiller: Yes, it is possible that the relation exists but no relation model object is available yet, so checking is necessary.

src/charm.py
```diff
@@ -252,6 +333,312 @@ def _on_nginx_route_broken(self, _: Any) -> None:
...
+            return
+        tls_rel_data = tls_certificates_relation.data[self.app]
+        if tls_rel_data[f"certificate-{hostname}"]:
+            event.set_results(
+                {
+                    f"certificate-{hostname}": tls_rel_data[f"certificate-{hostname}"],
+                    f"ca-{hostname}": tls_rel_data[f"ca-{hostname}"],
+                    f"chain-{hostname}": tls_rel_data[f"chain-{hostname}"],
+                }
+            )
```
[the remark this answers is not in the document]
gtrkiller: Sadly no, I have to check the relation for the data. The action is for testing and debugging purposes only, so there shouldn't be much of a problem I think.

src/charm.py
gtrkiller: Not really. the purpose of this method is to erase any remnants of the tls relation. The relation data will be deleted from the certificates relation whether we use secrets or not because the relation is extinct. If we do have secrets, however, we want to erase the existing secrets, completing the erase of the relevant data.

src/charm.py
gtrkiller: If we have secrets enabled, we currently store things in the secret and the tls relation databag, if we don't we just use the relation databag. So, the case is actually managed in this case. If what we want is to use only secrets and not write the relation databag when secrets are enabled, please let me know so I can make the changes.

src/charm.py
gtrkiller: See above

src/charm.py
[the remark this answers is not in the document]
gtrkiller: Only the certificate. For the CA and chain I must check the relation data.

src/charm.py
```diff
@@ -252,6 +333,312 @@ def _on_nginx_route_broken(self, _: Any) -> None:
...
+        tls_certificates_relation = self.model.get_relation("certificates")
+        if not tls_certificates_relation:
+            self.unit.status = WaitingStatus("Waiting for peer relation to be created")
+            event.defer()
+            return
+        relation = self._get_relation()
+        if relation is None:
+            self._cleanup()
+            self.unit.status = WaitingStatus("waiting for relation")
+            event.defer()
```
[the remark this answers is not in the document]
gtrkiller: After testing for some time, it is relevant. If we do not defer, it generates a race condition between generating the juju secret for the hostname and the hostname change when relating ingress to a requirer charm in some cases, and that can lead to juju secrets not being found or being created with a hostname that it is incorrect. Given this circumstance, I think deferring is the best choice.

src/controller.py
gregory-schiano: Well this PR is leading to this pylint warning, so that seems to be in the scope of that PR. I don't think this will require a big refactoring, but just moving each controller in a dedicated module. Does it sound ok ?

gtrkiller: Ok, I'll refactor in this PR then :+1:

src/charm.py
gregory-schiano: Well my idea is indeed to only store the secret ID in the relation databag when secrets are enabled, and fallback on peer relation otherwise. Concerning TLS relation that's a different story as the support should come from the library.
What do you think about this proposal ?

src/charm.py
gregory-schiano: Closing this comment, let's continue the discussion on the other thread

src/charm.py
gregory-schiano: Hmm I need to dig again, but I think I've read that even when a relation is removed and we get the event, there could be remnant of the data in the databag which could require explicitly flushing these data

src/charm.py
gtrkiller: Makes perfect sense. I'll implement it right away.

src/tls_relation.py
```diff
@@ -161,6 +161,7 @@ def certificate_relation_joined(  # type: ignore[no-untyped-def]
             charm: The Ingress charm that has the TLS relation
         """
         tls_certificates_relation = self.get_tls_relation(charm)
+        peer_relation = charm.model.get_relation("nginx-peers")
```
gregory-schiano, line 164: Could you use a constant for the peer relation name ?

src/tls_relation.py
```diff
@@ -207,16 +209,22 @@ def certificate_relation_created(  # type: ignore[no-untyped-def]
...
                 secret.set_content(private_key_dict)
+                secret_id = secret.id
             except SecretNotFoundError:
                 secret = charm.app.add_secret(
                     content=private_key_dict, label=f"private-key-{hostname}"
                 )
-        self.update_relation_data_fields(private_key_dict, tls_certificates_relation, charm.app)
-        peer_relation = charm.model.get_relation("nginx-peers")
-        self.update_relation_data_fields(private_key_dict, peer_relation, charm.app)
+                secret_id = secret.id
```
gregory-schiano, line 221: You probably don't need to repeat this line twice and directly use `secret.id` later on

gregory-schiano (commented):
Few minor comments

gregory-schiano (approved):
LGTM

src/controller/endpoint_slice.py
```diff
@@ -0,0 +1,169 @@
...
+        """
+        address_type = definition.upstream_endpoint_type
+        return kubernetes.client.V1EndpointSlice(
+            api_version="discovery.k8s.io/v1",
+            kind="EndpointSlice",
+            metadata=kubernetes.client.V1ObjectMeta(
+                name=definition.k8s_endpoint_slice_name,
+                namespace=definition.service_namespace,
+                labels={
+                    **self._labels,
```
amandahla, line 65: Will the "app.juju.is/created-by=nginx-ingress-integrator
                  nginx-ingress-integrator.charm.juju.is/managed-by=nginx-ingress-integrator" labels be included in here?

gtrkiller: yes, all those labels will be included in there

amandahla (approved):
LGTM

PR #102, opened 2023-11-08
Update charmcraft.yaml to declare arm64 support

weiiwang01:
The `nginx-ingress-integrator` charm is using pure Python, thereby it *should* compatibility with arm64 architectures. This pull request introduces the declarations to enable deployment on arm64 machines.


arturo-seijas (approved):
LGTM

jdkandersson (approved):

Issue #111, opened 2024-02-22
Add support for enable-access-log


alnvdl-work:
### Enhancement Proposal

For charms of applications that generate a lot of logs, disabling access logs with `enable-access-log: false` helps save significant amounts of storage. This is especially relevant if the application is already logging the requests it's receiving. We have some examples of this internally in Canonical.

weiiwang01:
I think this option is for the ingress controller and it will disable logging for the entire ingress controller? If so, I think this might be better managed by the cluster system administrator rather than being managed at the charm level.

weiiwang01:
@alnvdl-work that's right, sorry I misread the configuration. I agree this is a very good addition to the `nginx-ingress-integrator`!

alnvdl-work:
@weiiwang01 Is there a timeline for this being available? We depend on this for deploying some of our services that receive a lot of traffic.

weiiwang01:
> @weiiwang01 Is there a timeline for this being available? We depend on this for deploying some of our services that receive a lot of traffic.

@alnvdl-work Sorry, I have added it in this pull request: https://github.com/canonical/nginx-ingress-integrator-operator/pull/121 I will make it merged and released soon.

weiiwang01:
Revision 91 released adding the option `enable-access-log`.

PR #121, opened 2024-03-17
Add the disable-access-log option

weiiwang01:
Add a new option, `disable-access-log`, to control the `nginx.ingress.kubernetes.io/enable-access-log` annotation on ingress resources. 

Since this option is boolean value, flip the name to `disable-access-log` to ensure that the null value defaults to enabling the access log, which aligns with the default value of `nginx.ingress.kubernetes.io/enable-access-log` annotation.

fulfill: https://github.com/canonical/nginx-ingress-integrator-operator/issues/111

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -12,6 +12,10 @@ options:
       Indicates how NGINX should communicate with the backend service. Valid
       Values: HTTP, HTTPS, GRPC, GRPCS, AJP and FCGI.
     type: string
+  disable-access-log:
+    description: >
+      Setting this to true disables access log for the ingress using nginx.ingress.kubernetes.io/enable-access-log.
+    type: boolean
   ingress-class:
     default: ""
     description: |
--- a/lib/charms/nginx_ingress_integrator/v0/nginx_route.py
+++ b/lib/charms/nginx_ingress_integrator/v0/nginx_route.py
@@ -86,7 +86,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 5
+LIBPATCH = 6
 
 __all__ = ["require_nginx_route", "provide_nginx_route"]
 
@@ -181,6 +181,7 @@ def require_nginx_route(  # pylint: disable=too-many-locals,too-many-branches,to
     service_port: int,
     additional_hostnames: typing.Optional[str] = None,
     backend_protocol: typing.Optional[str] = None,
+    disable_access_log: bool = False,
     limit_rps: typing.Optional[int] = None,
     limit_whitelist: typing.Optional[str] = None,
     max_body_size: typing.Optional[int] = None,
@@ -211,6 +212,8 @@ def require_nginx_route(  # pylint: disable=too-many-locals,too-many-branches,to
             additional-hostnames option via relation, optional.
         backend_protocol: configure Nginx ingress integrator
             backend-protocol option via relation, optional.
+        disable_access_log: disable access logs with
+            nginx.ingress.kubernetes.io/enable-access-log option.
         limit_rps: configure Nginx ingress integrator limit-rps
             option via relation, optional.
         limit_whitelist: configure Nginx ingress integrator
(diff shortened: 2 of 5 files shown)


arturo-seijas (dismissed):

weiiwang01:
I'm still a little bit somewhat undecided about naming the new option `disable-access-log` instead of `enable-access-log`. On one hand, using `disable-access-log` aligns with programmer intuition, where a zero value or null value for a boolean type is false. On the other hand, we have a theme of matching the charm option with the actual annotation name here. What do you think? @arturo-seijas @alnvdl-work @mthaddon

alnvdl-work:
In ingress-nginx:
- `disable-access-log` is used as a controller-wide setting, and not an ingress setting: https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/configmap/#disable-access-log
`enable-access-log` is used as an ingress specific setting: https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#enable-access-log

This is already confusing enough on the K8s side, so there's no perfect solution.

So renaming `enable-access-log` to `disable-access-log` here would only add to the confusion of users. I'd argue for keeping the same name as ingress-nginx does: `enable-access-log`.

weiiwang01:
> In ingress-nginx:
> 
> * `disable-access-log` is used as a controller-wide setting, and not an ingress setting: https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/configmap/#disable-access-log
>   `enable-access-log` is used as an ingress specific setting: https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#enable-access-log
> 
> This is already confusing enough on the K8s side, so there's no perfect solution.
> 
> I think renaming `enable-access-log` to `disable-access-log` here would only add to the confusion of users. I'd argue for keeping the same name as ingress-nginx does: `enable-access-log`.

Yes, you are absolutely right; this is indeed very confusing. I have flipped this back to `enable-access-log`. thank you!

config.yaml
```diff
@@ -12,6 +12,10 @@ options:
       Indicates how NGINX should communicate with the backend service. Valid
       Values: HTTP, HTTPS, GRPC, GRPCS, AJP and FCGI.
     type: string
+  enable-access-log:
+    description: >
+      Setting this to false disables access log for the ingress using nginx.ingress.kubernetes.io/enable-access-log.
+    type: boolean
```
mthaddon, line 18: I think we might want to specify a default value of `true` to make it more obvious to someone reading this.

weiiwang01: The problem with having a default value in a boolean configuration is that the configuration value will always exist (either `True` or `False`), thereby always overriding the relational data and essentially prohibiting controlling this option via the relation. If there's no default value, then there will be three states (`True`, `False`, or `None`), allowing the relation to take control when the configuration is `None`.

How about adding this to the description?

> If this setting is not set anywhere (config or relation), it defaults to true.

weiiwang01: Added, thanks!

alnvdl-work: I don't see `default: true` here, or is `true` the default value if not informed?

weiiwang01: I added some explanations to the description, but I didn't change the default value. The value is `null` if not set.

The challenge of having a default boolean value is that, in the `nginx-ingress-integrator` configuration, it has higher precedence over the relation. If there's a default value for a boolean configuration, then the configuration is always defined, which means the configuration will always override the relation. This means that the relation cannot control this option.

However, we can modify the logic slightly and define disabling access logs as a "dominant gene", where either relation or config being false results in false. This approach would allow the relation to "override" the configuration when the relation data is false (*).

|                            | config is true  |  config is false  |
| --------------- | -------------- | --------------- | 
| relation is true  |       true            |        false          |
| relation is false |      false*           |        false          |
| relation unset   |       true            |        false          |

What do you think? By the way, below is the current truth table:

|                            | config is true  |  config is false  | config unset |
| --------------- | -------------- | --------------- | ------------ |
| relation is true  |       true            |        false            |      true         |
| relation is false |       true            |        false            |      false        |
| relation unset    |       true            |        false            |      true         |

alnvdl-work: Ahh, it makes sense now. I think keeping the current truth table you showed makes total sense, and it keeps backwards-compatibility. Thanks!

alnvdl-work (approved):
LGTM, thanks!

arturo-seijas (approved):

merkata (approved):

mthaddon (approved):
LGTM, thx

Issue #123, opened 2024-04-16
TLS certificate created via tls-certificates relation can't be decoded by nginx ingress controller


mthaddon:
### Bug Description

Currently we create a TLS certificate via the tls-certificates relation with a private key per https://github.com/canonical/nginx-ingress-integrator-operator/blob/main/src/tls_relation.py#L206. But we're not making that private key available to kubernetes, so the nginx-ingress controller can't decode the certificate, and falls back to the default certificate.

We should either remove the private key (traefik doesn't use one when it creates its certificate per https://github.com/canonical/traefik-k8s-operator/blob/565cd91d99545876cd4affeddd1055b9981791cb/lib/charms/observability_libs/v0/cert_handler.py#L218) or find a way to pass it to the nginx-ingress controller so that the certificate can be used

### To Reproduce

1. Deploy the charm and relate to httprequest-lego-k8s
2. Confirm the requested certificate can't be decoded by nginx-ingress controller

### Environment

Openstack, Juju 3.1.7.

### Relevant log output

```shell
# From the ingress controller logs
W0416 07:03:07.096578       6 backend_ssl.go:45] Error obtaining X.509 certificate: unexpected error creating SSL Cert: certificate and private key does not have a matching public key: tls: failed to parse private key
```

gregnebu:
After a quick analysis, the issue comes from the fact that the private keys is encrypted using a passphrase and stored in the secret encrypted. Since nginx ingress controller doesn't seem to support getting the passphrase to decrypt the private key, nginx-ingress-integrator should ensure that when the secret is created, the private key is stored decrypted.

Suggestion of fix:
- Since the tls-certificate library doesn't provide a way to decrypt a private key, I'd create a method within the `tls-reation` module to decrypt a given private key (which will require adding `cryptography` as a dependency
```python
from cryptography.hazmat.primitives import serialization
from cryptography.hazmat.primitives.asymmetric.rsa import RSAPrivateKey

def load_encrypted_private_key(encrypted_private_key: bytes, password: str) -> RSAPrivateKey:
    """
    Load and decrypt an RSA private key that is PEM-encoded and encrypted.
    
    Args:
        encrypted_private_key: The encrypted PEM file.
        password: Password used to encrypt the key.
    
    Returns:
        The decrypted RSA private key.
    """

    private_key = serialization.load_pem_private_key(
        encrypted_private_key,
        password=password
    )

    return private_key
```
[...]

PR #124, opened 2024-04-17
Decrypt private keys before storing them in secrets

gregory-schiano:
Related to issue: https://github.com/canonical/nginx-ingress-integrator-operator/issues/123

### Overview

In the current implementation when the K8S secret containing the private key is created, it stores the encrypted private key. This is unsupported by Nginx. Aim of this PR is to keep the private keys encrypted in secret/relation, but decrypt them before dumping them in K8S secrets

### Rationale

Since Nginx doesn't support providing the passphrase, that only solution is to store the private key decrypted. In order to keep them encrypted in the Juju secrets / peer relation, the private key is decrypted only when being dumped to K8S secrets

### Juju Events Changes

None

### Module Changes

* Added `get_decrypted_keys` to `TLSRelationService`.
* Did a bit of refactoring in `TLSRelationService`

### Library Changes

None

### Checklist

- [X] The [charm style guide](https://juju.is/docs/sdk/styleguide) was applied
- [X] The [contributing guide](https://github.com/canonical/is-charms-contributing-guide) was applied
- [X] The changes are compliant with [ISD054 - Managing Charm Complexity](https://discourse.charmhub.io/t/specification-isd014-managing-charm-complexity/11619)
- [X] The PR is tagged with appropriate label (`urgent`, `trivial`, `complex`)

Also fixed some issues found along the way:
* When the `AllCertificatesInvalidatedEvent` is triggered while removing the application, there aren't any relation available anymore and the charm was failing
* When adding the `tls-certificate` relation, the `ActiveStatus` message containing the IP address of the ingress was overriden
* Nginx ingress controller shipped with Microk8s doesn't support setting only the `Subject` in the certificate, it requires also a `SAN`


src/tls_relation.py
```diff
@@ -12,23 +12,31 @@
...
 
 from consts import PEER_RELATION_NAME, TLS_CERT
 
 
 class TLSRelationService:
     """TLS Relation service class."""
 
-    def __init__(self) -> None:
-        """Init method for the class."""
+    def __init__(self, model: Model, app: Application) -> None:
```
arturo-seijas, line 29: Isn't app the same as model.app?

gregory-schiano: Yes it is, let's say it's a first step of improvement compared to the fact that multiple methods so far were taking the whole charm (leading to a dependency with the charm in this module)

Currently I need the model and app for manage secrets, I'd like to find a solution for this case
model is also used to manage the relations but they might be passed to the constructor

arturo-seijas (commented):
It looks like you'll need to exclude tests/files/* from the license checks

gregory-schiano:
While manually testing the PR is found and solved several issues:
- When the `AllCertificatesInvalidatedEvent` is triggered while removing the application, there aren't any relation available anymore and the charm was failing
- When adding the tls-certificate relation, the `ActiveStatus` message containing the IP address of the ingress was overriden
- Nginx ingress controller shipped with Microk8s doesn't support setting only the `Subject` in the certificate, it requires also a `SAN`

src/tls_relation.py
```diff
@@ -244,73 +227,114 @@ def certificate_relation_available(  # type: ignore[no-untyped-def]
...
+            hostname: The hostname of the private key we want to fetch.
+
+        Returns:
+            The encrypted private key.
+        """
+        private_key_dict = {}
+        if JujuVersion.from_environ().has_secrets:
+            secret = self.charm_model.get_secret(label=f"private-key-{hostname}")
+            private_key_dict["key"] = secret.get_content()["key"]
+            private_key_dict["password"] = secret.get_content()["password"]
```
amandahla, line 330: If the key and the password are stored in the same place, why is the key encrypted?

gregory-schiano: That's true, I didn't change the implementation here, might be interesting to store them in separate secrets. There's unfortunately no good way to separate the password from the key, similarly the key is dumped into k8s unencrypted. But I'd rather have from the beginning an encrypted key, than manipulating non encrypted certificates.

mthaddon (approved):

src/tls_relation.py
```diff
@@ -138,70 +137,57 @@ def get_hostname_from_csr(self, tls_relation: Relation, app: Application, csr: s
...
-            ).encode()
+        assert isinstance(peer_relation, Relation)  # nosec
+        private_key_dict = self._get_private_key(hostname)
         csr = generate_csr(
-            private_key=private_key_dict["key"],
-            private_key_password=private_key_dict["password"],
+            private_key=private_key_dict["key"].encode(),
+            private_key_password=private_key_dict["password"].encode(),
             subject=hostname,
+            sans_dns=[hostname],
```
amandahla, line 170: Cool, it would be interesting to have a config for defining SANs in certificate.

amandahla (approved):
LGTM

src/tls_relation.py
gregory-schiano: Well not sure that would be relevant, we took the decision to issue 1 certificate per domain managed by nginx-ingress-integrator.
The main reasoning is that let's encrypt has pretty low rate limits in terms of certificate revocation, and when a domain is added or removed, if we have all the domains in the same certificate it requires to revoke it.
On the other hands 1 certificate per domain requires more certificates, but the rate limit is higher on let's encrypt.

The SAN here was added because recent versions of Nginx Ingress Controllers complains that Subject is deprecated and that we should use SAN

PR #126, opened 2024-04-22
Fix missing ingress url in ingress relation

codethulu:
Applicable spec: <link>

### Overview
This PR addresses the missing url field in the databag, and includes updates to tests to confirm this behaviour.

### Rationale

Currently, the url is missing from the ingress relation databag. This is data that is expected to be in the databag.

### Juju Events Changes

No changes.

### Module Changes

No significant changes, ingress definition now throws an error if path-routes are not prefixed with a `/`.

### Library Changes

N/A

### Checklist

- [x] The PR is tagged with appropriate label (`urgent`, `trivial`, `complex`)


codethulu:
need to ask question before merging

src/charm.py
```diff
@@ -319,6 +329,10 @@ def _update_ingress(self) -> None:
             ingress_controller = self._get_ingress_controller(namespace)
             ingress_ips = ingress_controller.get_ingress_ips()
             message = f"Ingress IP(s): {', '.join(ingress_ips)}" if ingress_ips else ""
+
+            if self.model.get_relation("ingress") is not None:
+                self._ingress_provider.publish_url(relation, ("http://" + hostnames[0]))
```
arturo-seijas, line 334: Will `hostnames` always contain an element at this point?

src/charm.py
```diff
@@ -228,6 +231,11 @@ def _check_precondition(self) -> None:
             raise InvalidIngressError(
                 "nginx-ingress-integrator cannot establish more than one relation at a time"
             )
+        if ingress_relation is not None and len(hostnames) > 1:
```
arturo-seijas, line 234: Why two hostnames and not one?

tests/integration/test_ingress_relation.py
```diff
@@ -55,13 +60,18 @@ def start_server(self):
         deploy_any_charm(json.dumps(src_overwrite)),
         build_and_deploy_ingress(),
     )
+
     await ingress.set_config({"service-hostname": "any"})
     await model.wait_for_idle()
-    await run_action("any", "rpc", method="start_server")
+    await asyncio.sleep(5)
```
arturo-seijas, line 66: If you need these `sleep` statements, there's probably something wrong with the charm. It should report ready status only when operational

codethulu: there seems to be some strange race condition at the moment, where the url only appears sometimes. the original integration test was broken initially, except it never checked the relation so it went un-noticed

src/charm.py
codethulu: we want to make sure that the number of hostnames is 1. if there are any more than 1 hostname (and we are using the ingress relation) then we should raise an error, so that the charm is blocked.

arturo-seijas: Gotcha, I was reading it wrong

src/charm.py
gregory-schiano: We should not hardcode `http` here, if the `tls-certificates` relation is set, we should be in `https`

gregory-schiano: Oh and the URL may contain a path too... definitely the way the URL is build should be better though !

tests/integration/test_ingress_relation.py
mthaddon: We discussed this IRL and the consensus was to update anycharm to go into Blocked status until it receives the info it needs over the relation. We should then be able to rely on `wait_for_idle`.

src/charm.py
```diff
@@ -319,10 +329,46 @@ def _update_ingress(self) -> None:
             ingress_controller = self._get_ingress_controller(namespace)
             ingress_ips = ingress_controller.get_ingress_ips()
             message = f"Ingress IP(s): {', '.join(ingress_ips)}" if ingress_ips else ""
+
+            if self.model.get_relation("ingress") is not None:
+                url = self._generate_ingress_url(hostnames) or ingress_ips[0]
```
mthaddon, line 334: `ingress_ips[0]` won't include the scheme or path, but it should (we should also have a unit test that confirms the case when we don't have a hostname).

src/charm.py
```diff
@@ -319,10 +329,46 @@ def _update_ingress(self) -> None:
...
+        )
+        prefix = "https" if tls_present else "http"
+
+        if not self.config.get("path-routes"):
+            return f"{prefix}://{hostname}"
+
+        pathroutes = self.config.get("path-routes", "").split(",")
+
+        if len(pathroutes) > 1:
+            raise InvalidIngressError("Ingress relation does not support multiple hostnames.")
```
mthaddon, line 369: The message here seems to be wrong?

tests/integration/test_ingress_relation.py
```diff
@@ -25,22 +25,31 @@ async def test_ingress_relation(
...
+    import ops
     from any_charm_base import AnyCharmBase
     from ingress import IngressPerAppRequirer
     class AnyCharm(AnyCharmBase):
         def __init__(self, *args, **kwargs):
             super().__init__(*args, **kwargs)
             self.ingress = IngressPerAppRequirer(self, port=8080)
+            self.unit.status = ops.BlockedStatus("Waiting for ingress relation")
+            self.framework.observe(
+                self.on.ingress_relation_changed, self._on_ingress_relation_joined
```
mthaddon, line 37: It's not really important, but it seems odd to be observing `ingress_relation_changed` but calling `_on_ingress_relation_joined`. Might be worth being consistent there.

codethulu: Ah my bad, was experimenting with different relation events and must have forgotten to rename the method

src/charm.py
codethulu: Fixed

src/charm.py
```diff
@@ -319,10 +329,46 @@ def _update_ingress(self) -> None:
...
+        return f"{prefix}://{hostname}{pathroutes[0]}"
```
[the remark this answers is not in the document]
codethulu: addressed

tests/integration/test_ingress_relation.py
```diff
@@ -55,13 +64,16 @@ def start_server(self):
...
-    await model.wait_for_idle()
+    await model.wait_for_idle(status="active")
+    await run_action("any", "rpc", method="start_server")
+
     response = requests.get(
         f"http://127.0.0.1/{model.name}-any/ok", headers={"Host": "any"}, timeout=5
     )
-    assert response.text == "ok"
+
+    assert response.text.rpartition("/")[0] == "http://any"
```
mthaddon, line 78: Have discussed this a bit outside this PR and it seems like we're doing this because we're seeing random path-routes values. We should try and figure out where those are coming from and why we can't set it to a specific value in this test.

src/charm.py
```diff
@@ -224,10 +224,16 @@ def _check_precondition(self) -> None:
             )
         nginx_route_relation = self.model.get_relation("nginx-route")
         ingress_relation = self.model.get_relation("ingress")
+        # The charm only supports one relation at a time, so we check if both are present
+        # and raise an error if they are.
         if nginx_route_relation is not None and ingress_relation is not None:
             raise InvalidIngressError(
                 "nginx-ingress-integrator cannot establish more than one relation at a time"
```
yanksyoon, line 231: I guess out of scope for this PR but I think the error message could be slightly better phrased as an actionable item: such as "Both Nginx relation and Ingress relation found. Please remove either one"

tests/integration/test_ingress_relation.py
```diff
@@ -55,13 +64,16 @@ def start_server(self):
         deploy_any_charm(json.dumps(src_overwrite)),
         build_and_deploy_ingress(),
     )
+
     await ingress.set_config({"service-hostname": "any"})
     await model.wait_for_idle()
-    await run_action("any", "rpc", method="start_server")
     await model.add_relation("any:ingress", "ingress:ingress")
-    await model.wait_for_idle()
+    await model.wait_for_idle(status="active")
```
yanksyoon, line 71: Up to you, but may I suggest `wait_for_active=True`?

yanksyoon (dismissed):
LGTM, minor suggestions thank you!

tests/integration/test_ingress_relation.py
codethulu: i don't have any strong feelings, is `wait_for_active` more conventional?

src/charm.py
codethulu: i can quickly make this change :)

tests/integration/test_ingress_relation.py
codethulu: using `status=active` because `wait_for_active=True` is being flagged as deprecated, and `status=active` is the suggested way.

src/charm.py
```diff
@@ -319,10 +326,41 @@ def _update_ingress(self) -> None:
             ingress_controller = self._get_ingress_controller(namespace)
             ingress_ips = ingress_controller.get_ingress_ips()
             message = f"Ingress IP(s): {', '.join(ingress_ips)}" if ingress_ips else ""
+
+            if self._get_definition_from_relation(relation).is_ingress_relation:
```
gregory-schiano, line 330: You already have the `definition` variable available (l314) so you can use it directly

src/charm.py
```diff
@@ -319,10 +326,41 @@ def _update_ingress(self) -> None:
...
+                hostnames = self.get_additional_hostnames()
+                url = self._generate_ingress_url(hostnames[0], definition.pathroutes)
```
gregory-schiano, line 332: Either add a comment specifying that there will always be an element available in the array point it out why, or add an `assert hostnames` to be defensive

src/charm.py
```diff
@@ -224,10 +224,16 @@ def _check_precondition(self) -> None:
...
         ingress_relation = self.model.get_relation("ingress")
+        # The charm only supports one relation at a time, so we check if both are present
+        # and raise an error if they are.
         if nginx_route_relation is not None and ingress_relation is not None:
             raise InvalidIngressError(
                 "nginx-ingress-integrator cannot establish more than one relation at a time"
             )
+        hostnames = self.get_additional_hostnames()
+        if ingress_relation is not None and len(hostnames) > 1:
+            self._ingress_provider.wipe_ingress_data(ingress_relation)
```
gregory-schiano, line 235: We're adding the `wipe_ingress_data` here and in `generate_ingress_url` when an `InvalidIngressError` is raised.
I'm wondering if we should have a more generic approach and call the wipe where we catch the `InvalidIngressError` (l336) ?

tox.ini
```diff
@@ -48,7 +48,7 @@ deps =
     pydocstyle>=2.10
     pylint
     pyproject-flake8<6.0.0
-    pytest
+    pytest==8.1.1
```
gregory-schiano, line 51: I think the issue is fixed now, we can remove the pinning

gregory-schiano (commented):
Small comments
I think we also need to add a test for `pathroutes` with routes that don't start with a `/`

src/charm.py
codethulu: if we do this, it will require a lot of extra logic since `relation` will be unbound, etc. we will have to have logic to check that the relation is set, that we are using ingress relation etc. this will include repetition of calling potentially unsafe in an `except` branch

gregory-schiano: Ok let's keep it as it is, but I'm wondering if we shouldn't systematically `wipe_ingress_data` when we have an `InvalidIngressError` whatever the issue is.

tox.ini
```diff
@@ -43,7 +43,7 @@ deps =
     flake8-docstrings-complete>=1.0.3
     flake8-test-docs>=1.0
     isort
-    mypy==1.6.1
+    mypy==1.6.1
```
nrobinaubertin, line 46: nitpick but this change could be removed

nrobinaubertin (approved):
Just one nitpick comment

tox.ini
codethulu: done

gregory-schiano (approved):
LGTM

mthaddon (approved):

PR #183, opened 2024-12-11
chore: bump tls lib to v4

gruyaume:
### Overview

Here we bump the TLS Certificates Library from v3 to v4. This change is a courtesy of the Credentials Charm Engineering team.

### Rationale

This change greatly reduces the amount of code needed to manage TLS Certificates. The TLS Certificates Library v4 takes care of most of the heavy lifting related to TLS Certificates. This new version of the library includes:

- Automated Private Key Management
- Automated CSR generation
- Automated Certificate management (including renewal)

With this new change, no need to directly manipulate TLS relation data in charm code, no need to observe most TLS related events.

### Juju Events Changes

- Stopped listening to most tls specific events as the lib takes care of certificate management

### Module Changes

- Removing the `src/tls_relation.py` abstraction as the tls lib takes care of most of what was in there.

### Library Changes

- tls_certificates: `v3` -> `v4`

### Integration tests changes

This project used to maintain its own dummy tls provider only used in the context of testing this charm. Instead of that here we moved to using the better maintained `self-signed-certificates` which serves the same purpose.

### Compatibility Notes
This new version of the library still uses the same version of the underlying charm relation interface, meaning that this charm here will remain compatible with the same TLS providers.

### Reference
- https://charmhub.io/tls-certificates-interface
- https://charmhub.io/self-signed-certificates

### Checklist

- [ ] The [charm style guide](https://juju.is/docs/sdk/styleguide) was applied
- [ ] The [contributing guide](https://github.com/canonical/is-charms-contributing-guide) was applied
- [ ] The changes are compliant with [ISD054 - Managing Charm Complexity](https://discourse.charmhub.io/t/specification-isd014-managing-charm-complexity/11619)
- [ ] The PR is tagged with appropriate label (`urgent`, `trivial`, `complex`)


gruyaume:
## A couple of notes on test failures

### Unit tests
Also, the unit test coverage test is failing because coverage went from 92 to 90%. That being said, the unit test coverage in `main.py` has remained at 96%. We removed one complete file (`src/tls_relation`) whose coverage was quite high - and caused the reduction in coverage. Here I didn't want to add dummy unit tests on code I didn't change. 

### Integration tests
I'm somewhat confused by the existing integration tests failure. Help from the maintainers here would be welcomed.

weiiwang01:
@gruyaume This is fantastic! Thank you for your awesome new tls charm library and your help with upgrading the `nginx-ingress-integrator`. I will review it tomorrow.

src/charm.py
```diff
@@ -395,21 +343,23 @@ def _on_get_certificate_action(self, event: ActionEvent) -> None:
...
-                    f"ca-{hostname}": tls_rel_data[f"ca-{hostname}"],
-                    f"chain-{hostname}": tls_rel_data[f"chain-{hostname}"],
-                }
-            )
-        else:
-            event.fail("Certificate not available")
+        provider_certificates, _ = self.certificates.get_assigned_certificates()
+        for provider_cert in provider_certificates:
+            if provider_cert.certificate.common_name == hostname:
+                provider_cert_json = provider_cert.to_json()
```
weiiwang01, line 353:
```suggestion
                provider_cert_json = json.loads(provider_cert.to_json())
```

tests/integration/test_cert_relation.py
```diff
@@ -85,16 +90,15 @@ async def build_and_deploy(
...
     await model.add_relation(f"{ANY_APP_NAME}:nginx-route", relation_name)
     await model.wait_for_idle()
-    provider_charm = await ops_test.build_charm(f"{PROVIDER_CHARM_DIR}/")
     await model.deploy(
-        provider_charm,
+        SELF_SIGNED_CERTIFICATES_CHARM_NAME,
         application_name=TLS_CERTIFICATES_PROVIDER_APP_NAME,
-        series="jammy",
+        channel="stable",
     )
```
weiiwang01, line 97: It seems that the `self-signed-certificates` on `latest/stable` is revision 155, which still uses version 3 of the TLS charm library. As a result, it doesn't recognize the new CSR in relation application data instead of unit data introduced in version 4 of the TLS charm library.

I assume this will also cause problems for anyone upgrading the `nginx-ingress-integrator` without upgrading TLS provider charms like the Lego provider?

tests/integration/test_cert_relation.py
```diff
@@ -178,13 +182,19 @@ async def test_given_additional_requirer_charm_deployed_when_relate_then_require
...
-    assert "ca-any" in action_output and action_output["ca-any"] is not None
-    assert "certificate-any" in action_output and action_output["certificate-any"] is not None
-    assert "chain-any" in action_output and action_output["chain-any"] is not None
+        keys = ["ca-any", "certificate-any", "chain-any"]
+        if action_output["return-code"] == 0 and all(action_output.get(key) for key in keys):
+            LOGGER.info("Certificate received")
+            break
+
+        LOGGER.info("Waiting for certificate")
+        time.sleep(5)
```
weiiwang01, line 197: Probably should raised an error when the timeout elapses here.

tests/integration/test_cert_relation.py
```diff
@@ -178,13 +182,19 @@ async def test_given_additional_requirer_charm_deployed_when_relate_then_require
     )
     requirer_unit = model.units[f"{new_requirer_app_name}/0"]
 
-    action = await requirer_unit.run_action(action_name="get-certificate", hostname="any")
+    t0 = time.time()
+    timeout = 600
+    while time.time() - t0 < timeout:
```
weiiwang01, line 187: The while loop here is probably unnecessary because we are calling `wait_for_idle` beforehand.

tests/integration/test_cert_relation.py
```diff
@@ -178,13 +182,19 @@ async def test_given_additional_requirer_charm_deployed_when_relate_then_require
...
+        keys = ["ca-any", "certificate-any", "chain-any"]
+        if action_output["return-code"] == 0 and all(action_output.get(key) for key in keys):
+            LOGGER.info("Certificate received")
+            break
+
+        LOGGER.info("Waiting for certificate")
+        time.sleep(5)
 
 
 @pytest.mark.usefixtures("build_and_deploy")
```
weiiwang01, line 200: We should probably remove the `test_given_enough_time_passed_then_certificate_expired` test, as the certificate signed by `self-signed-certificate` will not expire.

weiiwang01 (commented):
Overall looks good to me, thanks! Left some comments.

One more thing: maybe we should remove the old Juju secrets and data inside the relation created by the previous version of the `nginx-ingress-integrator` charm during the `upgrade-charm` hook. I did some testing, and the old secrets are all labeled with the pattern `private-key-{hostname}`, the old data in the relation application includes `ca-{hostname}`, `csr-{hostname}`, `certificate-{hostname}`, and `chain-{hostname}`.

weiiwang01:
> ## A couple of notes on test failures
> ### Unit tests
> 
> Also, the unit test coverage test is failing because coverage went from 92 to 90%. That being said, the unit test coverage in `main.py` has remained at 96%. We removed one complete file (`src/tls_relation`) whose coverage was quite high - and caused the reduction in coverage. Here I didn't want to add dummy unit tests on code I didn't change.
> ### Integration tests
> 
> ~~I'm somewhat confused by the existing integration tests failure. Help from the maintainers here would be welcomed. ~~ I believe I found the issue to be related to the fact we use a custom version of a TLS provider - here I replaced it with a real TLS provider which should provide more stable and realistic integration tests

Yes, please lower the coverage threshold.

src/charm.py
```diff
@@ -421,138 +371,57 @@ def get_all_hostnames(self) -> List[str]:
         relation = self._get_nginx_relation()  # type: ignore[arg-type]
         if not relation:
             return []
-        definition = self._get_definition_from_relation(relation)  # type: ignore[arg-type]
+        try:
+            definition = self._get_definition_from_relation(relation)  # type: ignore[arg-type]
+        except InvalidIngressError:
+            LOGGER.warning("Invalid ingress definition")
```
yanksyoon, line 377: I'm wondering if this should be an error (since for example in terraform we expected something with a relation but it did not get assigned whatever is expected)

yanksyoon (approved):
LGTM, minor questions, thank you :)

src/charm.py
```diff
@@ -421,138 +371,57 @@ def get_all_hostnames(self) -> List[str]:
...
-                    secret.remove_all_revisions()
-                except SecretNotFoundError:
-                    LOGGER.warning("Juju secret for %s already does not exist", hostname)
+    def _get_certificate_requests(self) -> List[str]:
+        """Get the certificate requests for the charm."""
+        hostnames = self.get_all_hostnames()
+        return [
+            CertificateRequestAttributes(common_name=hostname, sans_dns=frozenset([hostname]))
+            for hostname in hostnames
+        ]
```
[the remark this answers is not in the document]
gruyaume: It really depends on the use case, I'm not necessarily against it but I would not introduce this change part of this PR. The goal with this change here is a refactoring with no behavior changes.

src/charm.py
```diff
@@ -421,138 +371,57 @@ def get_all_hostnames(self) -> List[str]:
...
+
+    def _get_tls_certs(self) -> Dict[Union[str, None], Union[str, None]]:
+        """Get the TLS certificates provided by the TLS provider."""
+        certs: Dict[Union[str, None], Union[str, None]] = {}
+        if not hasattr(self, "certificates"):
+            return certs
```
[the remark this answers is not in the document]
gruyaume: `_get_tls_certs` is called through the `get_all_hostnames` function which is called in the init in order to get the hostnames for which to request certificates. 

Ideally it would be possible to list the hostnames for which we want to request certificates without providing certs/keys but I didn't want to make changes to `IngressDefinitionEssence` or anything Ingress related

tests/integration/test_cert_relation.py
gruyaume: Done

tests/integration/test_cert_relation.py
gruyaume: Agreed, done

tests/integration/test_cert_relation.py
gruyaume: I'm not sure honestly, as far as I can tell it's possible for the charm to be in active/idle status before it actually received the certificate

src/charm.py
gruyaume: Ok, I replaced the warning with an error

yhaliaw (approved):

weiiwang01:
@gruyaume , thanks again for the pull request. Overall, it looks good to me, and once the CI passes, we can proceed with merging it.

The integration test issue seems to be caused by `self-signed-certificates` on arm64, which doesn’t have a `beta` version, only an `edge` version. Changing the channel used in tests to `edge` should resolve this issue. Regarding the license headers check in the tests, you can use this script: https://gist.github.com/weiiwang01/1e25b871350b46b95fd85b0c18119998 to update all the licenses in the source files.

One small request before merging: could you please switch the `nginx-ingress-integrator` charm to use `Mode.UNIT` instead of `Mode.APP`? I believe `Mode.UNIT` should work with older versions of the tls certificates provider charm, avoiding a breaking change. We can introduce `Mode.APP` when we’re ready for the breaking change (perhaps in the next major release).

Thanks again for your help!

weiiwang01 (approved):

gruyaume:
> @gruyaume , thanks again for the pull request. Overall, it looks good to me, and once the CI passes, we can proceed with merging it.
> 
> The integration test issue seems to be caused by `self-signed-certificates` on arm64, which doesn’t have a `beta` version, only an `edge` version. Changing the channel used in tests to `edge` should resolve this issue. Regarding the license headers check in the tests, you can use this script: https://gist.github.com/weiiwang01/1e25b871350b46b95fd85b0c18119998 to update all the licenses in the source files.
> 
> One small request before merging: could you please switch the `nginx-ingress-integrator` charm to use `Mode.UNIT` instead of `Mode.APP`? I believe `Mode.UNIT` should work with older versions of the tls certificates provider charm, avoiding a breaking change. We can introduce `Mode.APP` when we’re ready for the breaking change (perhaps in the next major release).
> 
> Thanks again for your help!

Hello @weiiwang01 ,

Is there anything I can do to get this merged? :)

weiiwang01:
> > @gruyaume , thanks again for the pull request. Overall, it looks good to me, and once the CI passes, we can proceed with merging it.
> > The integration test issue seems to be caused by `self-signed-certificates` on arm64, which doesn’t have a `beta` version, only an `edge` version. Changing the channel used in tests to `edge` should resolve this issue. Regarding the license headers check in the tests, you can use this script: https://gist.github.com/weiiwang01/1e25b871350b46b95fd85b0c18119998 to update all the licenses in the source files.
> > One small request before merging: could you please switch the `nginx-ingress-integrator` charm to use `Mode.UNIT` instead of `Mode.APP`? I believe `Mode.UNIT` should work with older versions of the tls certificates provider charm, avoiding a breaking change. We can introduce `Mode.APP` when we’re ready for the breaking change (perhaps in the next major release).
> > Thanks again for your help!
> 
> Hello @weiiwang01 ,
> 
> Is there anything I can do to get this merged? :)

No, everything looks good to me. It's just that I am waiting for a second approval. Maybe @yanksyoon @yhaliaw you can help re-approve this pull request?

yanksyoon (approved):
Looks good on my side, thank you!

PR #192, opened 2025-03-03
Upgrade charm base and restore Juju 2 compatibility

weiiwang01:
Applicable spec: <link>

### Overview

Upgrade charm base to ubuntu@22.04 and restore juju 2 compatibility. Since Juju 2 doesn't support ubuntu 24.04, only upgrade the nginx-ingress-integrator to ubuntu 22.04 for now.

### Checklist

- [x] The documentation is generated using `src-docs`
- [x] The documentation for charmhub is updated
- [x] The PR is tagged with appropriate label (`urgent`, `trivial`, `complex`)


charmcraft.yaml
```diff
@@ -1,27 +1,196 @@
...
+config:
+  options:
+    additional-hostnames:
+      default: ""
+      description: >
+        Comma-separated list of additional hostnames for this ingress to listen on.
+      type: string
+    backend-protocol:
+      default: ""
+      description: >
```
seb4stien, line 59: Any specific reason to not use | as for other descriptions?

charmcraft.yaml
```diff
@@ -1,27 +1,196 @@
...
+        Indicates how NGINX should communicate with the backend service. Valid
+        Values: HTTP, HTTPS, GRPC, GRPCS, AJP and FCGI.
+      type: string
+    enable-access-log:
+      description: >-
```
seb4stien, line 64: Any specific reason to not use | as for other descriptions?

charmcraft.yaml
weiiwang01: I think for this instance, the new lines in the description are not actually intended as new lines in the description text but rather to break the long lines inside the yaml text. The configuration was added over a long period of time, so the style is not very consistent. I reviewed them and adjusted the scalar style accordingly. Thanks!

charmcraft.yaml
weiiwang01: See: https://github.com/canonical/nginx-ingress-integrator-operator/pull/192#discussion_r1977056275

weiiwang01:
The [style check test](https://github.com/canonical/nginx-ingress-integrator-operator/actions/runs/13626830054/job/38085753080?pr=192) seems to have failed due to the latest commit in https://github.com/canonical/praecepta/commit/a4c3f3b63b04d042a5014753d1b45af1b2a08f45. Will retry after it is reverted.

javierdelapuente (approved):

yanksyoon (approved):

yanksyoon:
Fixes #193 (referencing issues to keep track 🥷 )

Issue #229, opened 2025-07-07
Implement a `proxy-buffer-size` config option


edisile:
## Summary
K8s ingress definitions allow the usage of the `nginx.ingress.kubernetes.io/proxy-buffer-size` annotation, this charm should implement a config option that does the same

## Rationale
The default ingress configuration creates a 4KB headers proxy buffer, leading to 502 errors when the upstream server's response includes a header section that doesn't fit in the buffer. K8s allows to configure the buffer size, by using the `nginx.ingress.kubernetes.io/proxy-buffer-size` annotation on the ingress definition, allowing the user to select an appropriate value based on their needs.

We are currently facing this issue in our snapcraft.io Juju deployments: the `/login` call triggers a 302 response that carries a large `Location` header for the redirect to Ubuntu One SSO and a `Set-Cookie` header containing session data. When testing on a local environment we can access and modify the ingress definition manually by using `kubectl`, but this workaround isn't suited for a production environment.

I am currently working on a PR that implements the functionality described above, I will link it to this issue as soon as it's ready

PR #230, opened 2025-07-07
Add `proxy-buffer-size` config option

edisile:
### Overview

Added a config option allowing the use of `nginx.ingress.kubernetes.io/proxy-buffer-size` in ingress definition, as detailed in #229

### Rationale
Copy-paste from #229 for the sake of convenience:

> The default ingress configuration creates a 4KB headers proxy buffer, leading to 502 errors when the upstream server's response includes a header section that doesn't fit in the buffer. K8s allows to configure the buffer size, by using the `nginx.ingress.kubernetes.io/proxy-buffer-size` annotation on the ingress definition, allowing the user to select an appropriate value based on their needs.
>
> We are currently facing this issue in our snapcraft.io Juju deployments: the `/login` call triggers a 302 response that carries a large `Location` header for the redirect to Ubuntu One SSO and a `Set-Cookie` header containing session data. When testing on a local environment we can access and modify the ingress definition manually by using `kubectl`, but this workaround isn't suited for a production environment.

### Juju events changes
None

### Module changes
None

### Library changes
None

### Checklist

- [x] The documentation for charmhub is updated
- [ ] The PR is tagged with appropriate label (`urgent`, `trivial`, `senior-review-required`)
  - Not sure what the requirements are to apply each of the labels, I'd argue this is a `trivial` change that is also somewhat `urgent`
  - there should be no noticeable changes for users

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
 kubernetes==24.2.0 # last version with DiscoveryV1beta1Api
 ops==3.0.0
 jsonschema==4.24.0
-cryptography==45.0.4
+cryptography==45.0.5


srbouffard (approved):

weiiwang01 (approved):

lib/charms/nginx_ingress_integrator/v0/ingress.py
weiiwang01, line 1: You don't need to modify this charm library because `nginx_ingress_integrator.v0.ingress` is already deprecated. No new features will be added.

swetha1654 (approved):

src/ingress_definition.py
```diff
@@ -286,6 +286,12 @@ def owasp_modsecurity_custom_rules(self) -> str:
             str, self._get_config_or_relation_data("owasp-modsecurity-custom-rules", "")
         ).replace("\\n", "\n")
 
+    @property
+    def proxy_buffer_size(self) -> str:
```
gregory-schiano, line 290: It's recommended (as a good practice) to ensure that the proxy buffer size is a multiple of the memory page size. I'm wondering if we shouldn't at least send a warning when the configured value is not (or even enforce it)

edisile: Agreed, I'll add a warning for this case.
Also, I just realized that users could even set the value to 0 which would completely break the proxy, so I'll add some validation for this edge case and trigger an error

weiiwang01:
I have solved the problem in integration tests, it’s the same issue as in wordpress-k8s (pylibjuju version issue). I fixed the problem in another pull request and merged it into main.

@edisile can you merge main into your branch?

PR #245, opened 2025-08-19
feat: translate strip_prefix flag to ingress definition

florentianayuwono:
Applicable spec: None

### Overview

This PR provides corrected implementation of defining the value of `rewrite_enabled`, `rewrite_target`, and `path_routes` based on config, per-app ingress relation (`strip_prefix` flag) and nginx-route relation, in that order.

### Rationale

Currently there is a bug where the value is only defined by config and nginx-route relation, without considering the `strip_prefix` flag from per-app ingress relation.

### Juju events changes

None

### Module changes

None

### Library changes

None

### Checklist

- [x] The PR is tagged with appropriate label (`urgent`, `trivial`, `senior-review-required`)


src/ingress_definition.py
```diff
@@ -316,16 +316,54 @@ def proxy_connect_timeout(self) -> str:
 
     @property
     def rewrite_enabled(self) -> bool:
-        """Return whether rewriting should be enabled from config or relation."""
-        value = self._get_config_or_relation_data("rewrite-enabled", True)
+        """
+        Return whether rewriting should be enabled from config, per-app ingress
+        relation (strip_prefix), or nginx-route relation, in that order.
```
florentianayuwono, line 321: Hi @weiiwang01 wanted to clarify on this ordering priority: config -> ingress relation -> nginx-route relation when defining the value of `rewrite_enabled`, `rewrite_target`, and `path_routes`.

src/ingress_definition.py
```diff
@@ -316,16 +316,54 @@ def proxy_connect_timeout(self) -> str:
...
+        return False
 
     @property
     def rewrite_target(self) -> str:
-        """Return the rewrite target from config or relation."""
-        return cast(str, self._get_config_or_relation_data("rewrite-target", "/"))
+        """Return the rewrite target from config, per-app ingress
+        relation (strip_prefix), or nginx-route relation, in that order."""
+        value = self._get_config("rewrite-target")
+        if value is not None:
```
weiiwang01, line 350:
```suggestion
        if value:
```
I think we can't use `is` here because the default value of the `rewrite-target` configuration is `""`.

src/ingress_definition.py
```diff
@@ -440,14 +478,25 @@ def service_port(self) -> int:
     @property
     def path_routes(self) -> List[str]:
         """Return the path routes to use for the k8s ingress."""
+        value = self._get_config("path-routes")
+        if value is not None:
```
weiiwang01, line 482:
```suggestion
        if value:
```
I think we can't use `is` here because the default value of the `path-routes` configuration is `""`.

tests/integration/test_ingress_relation.py
```diff
@@ -74,3 +74,67 @@ def _on_ingress_relation_changed(self, event):
...
+        ),
+        "any_charm.py": any_charm_py,
+    }
+
+    _, ingress = await asyncio.gather(
+        deploy_any_charm(json.dumps(src_overwrite)),
+        build_and_deploy_ingress(),
+    )
+
+    await ingress.set_config({"service-hostname": "any", "path-routes": "/path(/|$)(.*)"})
```
weiiwang01, line 132: The `path-routes` is part of the integration functionality that needs to be tested, so I think we shouldn't set it in the charm configuration during tests.

tests/integration/test_ingress_relation.py
```diff
@@ -74,3 +74,67 @@ def _on_ingress_relation_changed(self, event):
...
+            file_path.parent.mkdir(exist_ok=True)
+            file_path.write_text(str(self.ingress.url))
+            proc_http = subprocess.Popen(
+                ["python3", "-m", "http.server", "-d", www_dir, "8080"],
+                start_new_session=True,
+            )
+
+        def _on_ingress_relation_changed(self, event):
+            self.unit.status = ops.ActiveStatus()
+    """
```
weiiwang01, line 117: This seems to be duplicated with the string in the `test_ingress_relation` function. Could you create a function that accepts `strip_prefix` as input and outputs the corresponding `any_charm` content to reduce code duplication?

src/ingress_definition.py
```diff
@@ -316,16 +316,54 @@ def proxy_connect_timeout(self) -> str:
...
+        """
+        Return whether rewriting should be enabled from config, per-app ingress
+        relation (strip_prefix), or nginx-route relation, in that order.
+        """
+        value = self._get_config("rewrite-enabled")
         # config data is typed, relation data is a string
         # Convert to string, then compare to a known value.
-        return str(value).lower() == "true"
+        if value is not None:
+            return str(value).lower() == "true"
```
weiiwang01, line 327: The default configuration value for the `rewrite-enabled` charm configuration is `false`, which means that `value is not None` will never be `True`.

One solution might be to remove the default value for `rewrite-enabled` in `charmcraft.yaml` so it defaults to `None`.

tests/integration/test_ingress_relation.py
```diff
@@ -74,3 +74,67 @@ def _on_ingress_relation_changed(self, event):
...
+        "ingress.py": pathlib.Path("lib/charms/traefik_k8s/v2/ingress.py").read_text(
+            encoding="utf-8"
+        ),
+        "any_charm.py": any_charm_py,
+    }
+
+    _, ingress = await asyncio.gather(
+        deploy_any_charm(json.dumps(src_overwrite)),
+        build_and_deploy_ingress(),
+    )
```
weiiwang01, line 130: Calling `deploy_any_charm` and `build_and_deploy_ingress` twice in the test (once in `test_ingress_relation` and once in `test_ingress_relation_strip_prefix`) will cause conflicts during tests, as they deploy charm applications with the same name inside the same Juju model.

I think the simplest solution is to merge the test in `test_ingress_relation_strip_prefix` into `test_ingress_relation`. For example, after the tests in `test_ingress_relation`, update the `src-overwrite` configuration value of `any-charm` to include `strip_prefix` and check if that changes the ingress as expected.

tests/integration/test_ingress_relation.py
florentianayuwono: I have updated the implementation to follow this. Can you help check if the idea is right? I've been waiting for the local integration test to run but seems like it's now stuck on "Waiting for model"

src/ingress_definition.py
```diff
@@ -316,16 +316,54 @@ def proxy_connect_timeout(self) -> str:
...
+        if self.is_ingress_relation:
+            try:
+                strip_prefix = (
+                    cast(IngressPerAppProvider, self.ingress_provider)
+                    .get_data(self.relation)
+                    .app.strip_prefix
+                )
+                if strip_prefix:
+                    return "/$2"  # matches regex capture group in path_route
+                return "/"
```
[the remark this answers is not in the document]
gregory-schiano: Shouldn't it be the contrary ?
When strip prefix is enabled we want that when `/my-model-my-app` is called on nginx, the prefix is removed and the backend is called on `/``
And when strip prefix is disabled, we want that `/my-model-my-app` is not rewriten and the backend is called on `/my-model-myapp`

src/ingress_definition.py
```diff
@@ -317,15 +317,49 @@ def proxy_connect_timeout(self) -> str:
...
+                return (
+                    cast(IngressPerAppProvider, self.ingress_provider)
+                    .get_data(self.relation)
+                    .app.strip_prefix
+                )
+            except DataValidationError as exc:
+                raise InvalidIngressError(msg=f"{exc}, cause: {exc.__cause__!r}") from exc
+
+        value = self._get_relation("rewrite-enabled")
+        if value:
```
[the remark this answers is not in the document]
florentianayuwono: I'm afraid that `_get_relation` will return either empty string or None if the relation doesn't exist. In the case of empty string, `if value is not None` will return `True`. But anyway even if it pass the conditional check it will still return False because of the `str(value).lower() == "true"` so yes `if value is not None` will work too. But using `if value` will immediately fail the check in case of empty string and immediately return `False`.

src/ingress_definition.py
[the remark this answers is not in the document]
florentianayuwono: Hmm is this assumption correct:
- when `strip_prefix=True`, the expected output should be something like:
```
juju config nginx-ingress-integrator path-routes='/documentation(/|$)(.*)'
juju config nginx-ingress-integrator rewrite-enabled='True'
juju config nginx-ingress-integrator rewrite-target='/$2'
```
- when `strip_prefix=False`, this is the current behaviour when rewrite-target simply becomes "/"

src/ingress_definition.py
```diff
@@ -317,15 +317,49 @@ def proxy_connect_timeout(self) -> str:
...
+                    .get_data(self.relation)
+                    .app.strip_prefix
+                )
+                if strip_prefix:
+                    return "/$2"  # matches regex capture group in path_route
+                return "/"
+            except DataValidationError as exc:
+                raise InvalidIngressError(msg=f"{exc}, cause: {exc.__cause__!r}") from exc
+
+        return cast(str, self._get_relation_data_or_config("rewrite-target", "/"))
```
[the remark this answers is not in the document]
florentianayuwono: We can, it's just that I want to minimize the lines. My assumption is that:
- In case of `_get_relation` exists -> return the value
- In case of it doesn't exist -> return "/"
- Since we already check that `_get_config` doesn't exist previously, `_get_relation_data_or_config` is essentially `_get_relation` with extra capability to inject the fallback value "/" immediately.

charmcraft.yaml
```diff
@@ -153,7 +153,6 @@ config:
         will be used if this config option is set to an empty value.
       type: string
     rewrite-enabled:
-      default: false
```
yhaliaw, line 156: Is removing the default intended?
Since we generally want to have sensible defaults.

docs/explanation/what-is-ingress.md
```diff
@@ -16,8 +16,9 @@ The charm can be configured via a relation, or via `juju config` directly. See t
 The reason for offering both relation and direct `juju config` support is that providing the relation means charm authors can make the experience better for end users by implementing the relation. If a charm doesn’t implement the relation it can still be used with this charm and configured manually.
 
 Using the relation, the charm supports the following features:
-
+<!-- vale off -->
```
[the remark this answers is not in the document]
florentianayuwono: It was temporarily excluded to pass the check! But now erin's new vale PR has fixed this :)

tests/integration/test_ingress_relation.py
```diff
@@ -8,69 +8,109 @@
...
+def make_any_charm_source(strip_prefix: bool = False) -> str:
+    """Generate the source code for the any-charm with ingress relation.
+
+    Args:
+        strip_prefix: Whether to strip the prefix from the URL path.
+
+    Return:
+        str: The source code of the any-charm.
     """
+    file_path_expr = 'www_dir / "testing-any" / "ok"' if not strip_prefix else 'www_dir / "ok"'
```
weiiwang01, line 25: The `testing` is dynamic based on the Juju model name. I think this should be a variable.

PR #250, opened 2025-08-29
feat: add use-regex

florentianayuwono:
Applicable spec: <link>

### Checklist

- [ ] The [charm style guide](https://juju.is/docs/sdk/styleguide) was applied
- [ ] The [contributing guide](https://github.com/canonical/is-charms-contributing-guide) was applied
- [ ] The changes are compliant with [ISD054 - Managing Charm Complexity](https://discourse.charmhub.io/t/specification-isd014-managing-charm-complexity/11619)
- [ ] The PR is tagged with appropriate label (`urgent`, `trivial`, `senior-review-required`)

### DIFF
--- a/src/controller/ingress.py
+++ b/src/controller/ingress.py
@@ -112,7 +112,7 @@ class IngressController(
         ingress_paths = [
             kubernetes.client.V1HTTPIngressPath(
                 path=path,
-                path_type="Prefix",
+                path_type="ImplementationSpecific" if definition.use_regex else "Prefix",
                 backend=kubernetes.client.V1IngressBackend(
                     service=kubernetes.client.V1IngressServiceBackend(
                         name=definition.k8s_service_name,
@@ -166,6 +166,8 @@ class IngressController(
             )
         if definition.rewrite_enabled:
             annotations["nginx.ingress.kubernetes.io/rewrite-target"] = definition.rewrite_target
+        if definition.use_regex:
+            annotations["nginx.ingress.kubernetes.io/use-regex"] = "true"
         if definition.session_cookie_max_age:
             annotations["nginx.ingress.kubernetes.io/affinity"] = "cookie"
             annotations["nginx.ingress.kubernetes.io/affinity-mode"] = "balanced"
--- a/src/ingress_definition.py
+++ b/src/ingress_definition.py
@@ -374,6 +374,31 @@ class IngressDefinitionEssence:  # pylint: disable=too-many-public-methods
         return cast(str, self._get_relation("rewrite-target", "/"))
 
     @property
+    def use_regex(self) -> bool:
+        """Return whether regex matching should be enabled."""
+        value = self._get_config("use-regex")
+        if value is not None:
+            return str(value).lower() == "true"
+
+        if self.is_ingress_relation:
+            try:
+                strip_prefix = (
+                    cast(IngressPerAppProvider, self.ingress_provider)
+                    .get_data(self.relation)
+                    .app.strip_prefix
+                )
+                if strip_prefix:
+                    return True
+            except DataValidationError as exc:
+                raise InvalidIngressError(msg=f"{exc}, cause: {exc.__cause__!r}") from exc
+
+        if self.rewrite_enabled and "$" in self.rewrite_target:
+            return True
+
+        value = self._get_relation("use-regex")
+        return value is not None and str(value).lower() == "true"
+
+    @property
     def service_namespace(self) -> str:
         """Return the namespace to operate on."""
         if self.is_ingress_relation:
@@ -634,6 +659,7 @@ class IngressDefinition:  # pylint: disable=too-many-public-methods,too-many-ins
     retry_errors: str
     rewrite_enabled: bool
     rewrite_target: str
+    use_regex: bool
     service_hostname: str
     service_name: str
     service_namespace: str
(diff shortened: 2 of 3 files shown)


src/ingress_definition.py
```diff
@@ -373,6 +373,31 @@ def rewrite_target(self) -> str:
...
+                    .get_data(self.relation)
+                    .app.strip_prefix
+                )
+                if strip_prefix:
+                    return True
+            except DataValidationError as exc:
+                raise InvalidIngressError(msg=f"{exc}, cause: {exc.__cause__!r}") from exc
+
+        if self.rewrite_enabled and "$" in self.rewrite_target:
+            return True
```
weiiwang01, line 396: Regex containing `$` is only a fraction of the cases where regex is used. I think this is a bit redundant.

src/controller/ingress.py
```diff
@@ -112,7 +112,7 @@ def _gen_resource_from_definition(  # noqa: C901
         ingress_paths = [
             kubernetes.client.V1HTTPIngressPath(
                 path=path,
-                path_type="Prefix",
+                path_type="ImplementationSpecific" if definition.use_regex else "Prefix",
```
arturo-seijas, line 115: Can you define the types in an enum?

src/ingress_definition.py
```diff
@@ -373,6 +373,31 @@ def rewrite_target(self) -> str:
...
+            try:
+                strip_prefix = (
+                    cast(IngressPerAppProvider, self.ingress_provider)
+                    .get_data(self.relation)
+                    .app.strip_prefix
+                )
+                if strip_prefix:
+                    return True
+            except DataValidationError as exc:
+                raise InvalidIngressError(msg=f"{exc}, cause: {exc.__cause__!r}") from exc
```
arturo-seijas, line 393: I think `raise InvalidIngressError() from exc` might be informative enough

src/ingress_definition.py
florentianayuwono: I see, so decide the value based on `rewrite_enabled`.

src/ingress_definition.py
```diff
@@ -373,6 +373,31 @@ def rewrite_target(self) -> str:
 
         return cast(str, self._get_relation("rewrite-target", "/"))
 
+    @property
+    def use_regex(self) -> bool:
```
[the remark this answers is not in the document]
florentianayuwono: Good suggestion!

src/ingress_definition.py
```diff
@@ -373,6 +373,31 @@ def rewrite_target(self) -> str:
...
+        """Return whether regex matching should be enabled."""
+        value = self._get_config("use-regex")
```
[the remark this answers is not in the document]
florentianayuwono: Oh I was trying to standardize with the way other functions also fetch the config.

src/ingress_definition.py
florentianayuwono: @weiiwang01 on second thought, I think we still need to detect whether regex is used by looking at the `rewrite_target`. Currently it can be done because `$` is used as a capture group, for example `/$2`. I am not sure if we should only rely on `rewrite_enabled` as there is Static Rewrite case where `rewrite_enabled=True`, and `rewrite_target` is a literal path without `$N`.
Example:
`rewrite-enabled=True` and `rewrite-target="/"` is static (all requests go to `/`) means that `use-regex=False` and `pathType=Prefix`.

src/ingress_definition.py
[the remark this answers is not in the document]
florentianayuwono: @arturo-seijas hmm I think I'm having trouble on implementing this because I still need to define `annotations["nginx.ingress.kubernetes.io/use-regex"]` which is currently done in `/controller/ingress.py` based on the Ingress Definition `use-regex`. Returning `pathType` here means that I still need to deduce the value of `annotations["nginx.ingress.kubernetes.io/use-regex"]` based on it, which imo seems a bit more complex than the current implementation.
```
if definition.use_regex:
            annotations["nginx.ingress.kubernetes.io/use-regex"] = "true"
```

src/ingress_definition.py
weiiwang01: Yes, what I suggested in the old comment is to simply not decide the `use-regex` value based on `rewrite-enabled` or `rewrite-target` at all. We can just set `use-regex` only when `use-regex` or `strip-prefix` is specified in the configuration or relation.

florentianayuwono: But then what about the case of Static/Dynamic Rewrite where `rewrite_enabled` is both set to `True`? We still need to check that if `rewrite_target="/"` then `use-regex=False`, and when `rewrite-target="/$2"` then `use-regex=True`.

src/ingress_definition.py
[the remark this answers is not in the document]
arturo-seijas: Let's keep it as it is, then

weiiwang01 (approved):

src/ingress_definition.py
florentianayuwono: I see, so right now we should scrap this part completely, and determine the `use-regex` value solely from the config/relation value of `use-regex` or `strip-prefix`, right?
```
        if self.rewrite_enabled and "$" in self.rewrite_target:
            return True
```

weiiwang01: Yes, if we can be 100% sure that the user is using regex, then I think it's fair to detect the regex. But realistically we can’t do the detection, then changing the use-regex value based on the rewrite-target input feels unpredictable and may be confusing to the user.

PR #418, opened 2026-06-02
ci: migrate integration tests and publishing to canonical/charm-ci

javierdelapuente:
## Overview
Replaces `canonical/operator-workflows` reusable workflows for integration tests and publishing with `canonical/charm-ci` (`opcli`).

### New files
| File | Purpose |
|---|---|
| `artifacts.yaml` | Declares the charm to build (no rocks/OCI images) |
| `spread.yaml` | Two integration suites: juju 3.6 amd64/arm64 and juju 2.9 amd64 |
| `concierge-juju3-microk8s.yaml` | Provisions juju 3.6/stable + microk8s 1.34-strict/stable |
| `concierge-juju2-microk8s.yaml` | Provisions juju 2.9/stable + microk8s 1.24/stable |

### Updated workflows
- **`integration_test.yaml`**: replaced with charm-ci `integration-test.yml`. arm64 is a backend in `spread.yaml`. Permissions moved to job level.
- **`publish_charm.yaml`**: replaced with charm-ci `publish-artifacts.yml`. arm64 merged in. Permissions moved to job level.
- **`comment.yaml`**: permissions moved to job level.

### Deleted workflows
- `integration_test_arm64.yaml` — covered by the `ubuntu-24.04-juju3-arm64` backend in `spread.yaml`
- `publish_charm_arm64.yaml` — merged into `publish_charm.yaml`

### Integration test changes
- **Removed `pytest-jubilant`**: replaced with a simple session-scoped `juju` fixture using `jubilant.temp_model` directly. Supports `--model`, `--keep-models`, and `--use-existing` options.
- **Migrated to `jubilant-backports`**: replaces `jubilant` so the same test code runs across Juju 2.9–4.x. Import changed to `import jubilant_backports as jubilant`.
- **Merged `integration-juju2` dep group**: now uses the same `integration` group (jubilant-backports supports both juju versions).
- **arm64 arch**: `PYTEST_ADDOPTS=--model-arch=arm64` is passed via `pytest-environment-template` in `spread.yaml` for arm64 suites; the juju fixture sets model constraints accordingly.

### Deleted workflows
- `integration_test_arm64.yaml`
- `publish_charm_arm64.yaml`

### Unchanged workflows
`test.yaml`, `promote_charm.yaml`, `auto_update_libs.yaml` — charm-ci has no equivalent for these.

## Notable trade-offs
- **Allure reports dropped**: charm-ci has no equivalent; the Saturday scheduled run will no longer produce an Allure report.

## Juju Events Changes
None.

## Module Changes
None.

## Library Changes
None.

## Checklist
- [x] `CHARMHUB_TOKEN` secret must be configured in the repository
- [x] Concierge MicroK8s addon names verified for targeted channels


artifacts.yaml
```diff
@@ -0,0 +1,11 @@
...
+# See LICENSE file for licensing details.
+
+version: 1
+charms:
+  - name: nginx-ingress-integrator
+    charmcraft-yaml: charmcraft.yaml
+    platforms:
+      - arch: amd64
+      - arch: arm64
+        runner: [ubuntu-24.04-arm]
```
seb4stien, line 11: Why do you have to put it here? I also see it in the spread.yaml, shouldn't it be in only one of them?

javierdelapuente: This is to build the charm (it has to be arm as there is no "cross compiling" for charms).

In spread.yaml is to run the integration tests.

spread.yaml
```diff
@@ -0,0 +1,54 @@
...
+            TOX_ENV: integration-juju2
+
+environment:
+  OPCLI_GIT_REF: '$(HOST: echo "${OPCLI_GIT_REF:-main}")'
+
+exclude:
+  - .git
+  - .tox
+  - .venv
+  - .*_cache
```
seb4stien, line 33: Do you need the "build/"?
Maybe it's too much, shouldn't we exclude "*.charm" and "*.rock" in case they have been built locally?

seb4stien (approved):

spread.yaml
```diff
@@ -0,0 +1,54 @@
...
+
+integration-suites:
+  tests/integration/:
+    working-dir: ./
+    summary: integration tests
+    systems: [ubuntu-24.04-juju3, ubuntu-24.04-juju3-arm64]
+    backends:
+      - integration-test
+    pytest-environment-template: |
+      {% if arch == "arm64" %}PYTEST_ADDOPTS=--model-arch=arm64{% endif %}
```
javierdelapuente, line 43: I think this is not needed, as the `testing` model created by concierge already has the correct model constraint.

spread.yaml
javierdelapuente: I think it is needed, as the expansion happens before the spread run, and I think spread needs the task.yaml inside (I have to review it anyway)

artifacts.yaml
javierdelapuente: I expect problems in the future with this (I mean, self hosted runners for things like the discourse rock that is very resource intensive)

javierdelapuente:
This is old. Close and redo.

